PR #14, opened 2022-02-03
Add Docker files and scripts

gshiroma:
This PR adds Docker files (`DockerFile`, `requirements.txt`, and `requirements.txt.forge`) and scripts `build_docker_image.sh`  and `load_docker_tar.sh` to the main repository.


rtburns-jpl:
Is the CentOS 8 base image a requirement? We could reduce the size a lot by using e.g. Alpine. Also, we might be able to save some size by using the system python instead of conda.

hfattahi:
> Is the CentOS 8 base image a requirement? We could reduce the size a lot by using e.g. Alpine. Also, we might be able to save some size by using the system python instead of conda.

Maybe @collinss-jpl or @LucaCinquini or @riverma can comment on this.

LucaCinquini:
I don't think Centos-8 is a requirement - in-fact, our PCM AMIs will be based on Oracle Linux 8. I think in general the OS needs to be on the NASA approved list - I am not sure Alpine is. I asked Susan Neely for the official list.

LucaCinquini:
Please see here list from Susan: https://cset.nasa.gov/operating-systems/
(must login via NASA ID)
and also:
https://cset.nasa.gov/ascs/frequently-asked-ascs-questions/#Linux-Why-are-only-some-Linux-distributions-Agency-approved

docker/Dockerfile
```diff
@@ -0,0 +1,75 @@
+FROM centos:8
+# FROM python:3.8 
+# substitute centos:8 with oracle linux 8 (ask JPL Susan)
```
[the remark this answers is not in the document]
gshiroma: Thank you for your feedback @rtburns-jpl . You said it's not necessary to decide on a base image now, but I guess it would be better to move now to a NASA-approved base image now, right? Dependencies requirements for this tool should be easy to meet for all base images. What would you suggest to use? Ubuntu 20.04?

LucaCinquini:
What is Nisar doing? Shall we follow their example?

gshiroma:
> 

Good question, @LucaCinquini . @rtburns-jpl can answer it better, but by looking at the ISCE3 repository, it seems that it's currently using `centos:8`,  `ubuntu:19.04`, and other base images for GPU (`nvidia/cuda:10.1-base`, `nvidia/cuda:9.2-runtime-centos7`).

docker/Dockerfile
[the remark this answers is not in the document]
hfattahi: I agree with @gshiroma let's make a decision now and be done with it.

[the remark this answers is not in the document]
rtburns-jpl: Okay, if we want to make a decision now I'd say Ubuntu 20.04 is the best option.

riverma:
@gshiroma @hfattahi I'm a little confused - how does this Docker image differ with the PGE Docker infrastructure @collinss-jpl is setting up? See his Docker build template [here](https://github.com/nasa/opera-sds-pge/blob/94d45c4dc0154153bcd40831dedd7b285895a7d2/.ci/docker/Dockerfile_dswx_hls).

gshiroma:
> 



> @gshiroma @hfattahi I'm a little confused - how does this Docker image differ in terms of goals with the PGE Docker infrastructure @collinss-jpl is setting up? See his Docker build template [here](https://github.com/nasa/opera-sds-pge/blob/94d45c4dc0154153bcd40831dedd7b285895a7d2/.ci/docker/Dockerfile_dswx_hls) for reference.

Hi @riverma , the PGE should wrap around/use this SAS. Considering the PGE Docker file that you pointed out, I think it loads the DSWx-HLS SAS image in line 17 and it executes the DSWx-HLS SAS in line 63?

riverma:
Hi 

> > @gshiroma @hfattahi I'm a little confused - how does this Docker image differ in terms of goals with the PGE Docker infrastructure @collinss-jpl is setting up? See his Docker build template [here](https://github.com/nasa/opera-sds-pge/blob/94d45c4dc0154153bcd40831dedd7b285895a7d2/.ci/docker/Dockerfile_dswx_hls) for reference.
> 
> Hi @riverma , the PGE should wrap around/use this SAS. Considering the PGE Docker file that you pointed out, I think it loads the DSWx-HLS SAS image in line 17 and it executes the DSWx-HLS SAS in line 63?

Hi @gshiroma - oh I see - this PR refers to the core SAS image DSWx-HLS PGE is supposed to wrap. @collinss-jpl does that sound right? Just wanted to ensure we're not duplicating extra work on either side.

collinss-jpl:
> Hi
> 
> > > @gshiroma @hfattahi I'm a little confused - how does this Docker image differ in terms of goals with the PGE Docker infrastructure @collinss-jpl is setting up? See his Docker build template [here](https://github.com/nasa/opera-sds-pge/blob/94d45c4dc0154153bcd40831dedd7b285895a7d2/.ci/docker/Dockerfile_dswx_hls) for reference.
> > 
> > 
> > Hi @riverma , the PGE should wrap around/use this SAS. Considering the PGE Docker file that you pointed out, I think it loads the DSWx-HLS SAS image in line 17 and it executes the DSWx-HLS SAS in line 63?
> 
> Hi @gshiroma - oh I see - this PR refers to the core SAS image DSWx-HLS PGE is supposed to wrap. @collinss-jpl does that sound right? Just wanted to ensure we're not duplicating extra work on either side.

@riverma That is correct

docker/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+# ENTRYPOINT [ "/opt/docker/bin/entrypoint" ]
+
+USER root
+RUN mkdir -p DSWx-HLS-0.1/build
+RUN export PATH=/opt/conda/bin:$PATH && cd DSWx-HLS-0.1 && python3 setup.py install
+
+RUN mkdir -p runconfig && chown conda:conda runconfig
+RUN mkdir -p input_dir && chown conda:conda input_dir
+RUN mkdir -p output_dir && chown conda:conda output_dir
+RUN mkdir -p scratch_dir && chown conda:conda scratch_dir
```
collinss-jpl, line 60: @gshiroma We may need to rethink this line, as its been the source of some permission issues when running the combined PGE/SAS image.

Typically, when running a Docker image that will be mounting Docker volumes, you specify that the container should run as the current user/group of the host machine (via the `-u $UID:$(id -g)` argument in the `docker run` call). We also provide docker volume mounts to the `runconfig`, `input_dir` and `output_dir` directories within the container, which in effect changes the owner/group from `conda:conda` to `$UID:$(id -g)`. This ensures that permissions for both input and output products are set correctly "on both sides" of the container. 

However, this still leaves `scratch_dir`, which since we don't mount anything to it, it is still owned by `conda:conda`. This has caused permissions errors when the PGE (running as `$UID:$(id -g)`) tries to write out a SAS-only RunConfig to feed to the Python executable. Either the PGE or SAS (or both) should be responsible for creating `scratch_dir` at runtime based on how the RunConfig is set, or the Dockerfile needs to fully open permissions so any user/group running the container can write to it.

Personally I'm for the former option of creating it dynamically, since the PGE already has logic to do this. Let me know y'alls thoughts on this.

gshiroma: Thank you, @collinss-jpl . Good point. I updated the directory permissions on commit [b10f426](https://github.com/opera-adt/PROTEUS/pull/14/commits/b10f426bd5658729b37c44efbd38229e1ca3ec33) substituting `conda:conda` with `$UID:$(id -g)` following your suggestion. I'm doing the same for `scratch_dir` which can also be mounted and handled the same way as the other directories. Would that address your concerns?

collinss-jpl: Hi @gshiroma, thanks for taking a look. Unfortunately this approach does not address this issue, as changing ownership to `$UID:$(id -g)` from *within* the Dockerfile would just set the user and owner to `root` (since line 53 of the Dockerfile changes the user to `root` before the directories are created).

When I talk about "opening up" the permissions, I meant something more like this:

`RUN mkdir -p scratch_dir && chown conda:conda scratch_dir && chmod 777 scratch_dir`

So basically the directory is created with fully open permissions for any user/group that runs the container. This should only need to be done for `scratch_dir`, since we mount the other directories as Docker volumes.

gshiroma: Thank you again, @collinss-jpl . I accepted your suggestions to open up `scratch_dir` permissions to all groups and users. I also reverted changes to other directories. Please, let me know if that's correct now.

gshiroma:
Hi @rtburns-jpl , I fixed the package name (from DSWx-HLS to proteus) in the Dockerfile. Please, let me know if you can build and load the Docker image now. Thanks.

docker/Dockerfile
collinss-jpl: Looks great, thanks for the prompt response

collinss-jpl (approved):

rtburns-jpl:
Thanks Gustavo, I was able to build & run the docker images with your updates. This LGTM

gshiroma:
> Thanks Gustavo, I was able to build & run the docker images with your updates. This LGTM

Thank you, @rtburns-jpl . Would you please approve the PR?

rtburns-jpl:
Oh haha, didn't see that I need to formally approve to unblock merging.

rtburns-jpl (approved):

hfattahi:
@collinss-jpl you will be using the Docker down the line. It would be great to make sure this PR addresses your needs. Could you review and approve if all looks good to you?

gshiroma:
> Oh haha, didn't see that I need to formally approve to unblock merging.

Thank you, @rtburns-jpl !

gshiroma:
> @collinss-jpl you will be using the Docker down the line. It would be great to make sure this PR addresses your needs. Could you review and approve if all looks good to you?

Yes, @collinss-jpl had previously approved this PR, but since I pushed changes to it, I think GitHub automatically removed @collinss-jpl 's approval.  @collinss-jpl , would you please take a look at the recent updates and let us know if you could approve this PR or if you have any concerns? Thank you.

collinss-jpl (approved):
Latest changes look good to me

gshiroma:
> Latest changes look good to me, but it looks like my approval doesn't count towards the merge requirement because I don't have write access to this repo, just FYI

Sounds good, @collinss-jpl . Thank you for reviewing this PR and its recent updates!

PR #18, opened 2022-04-20
Update DSWx-HLS for point release in mid May

gshiroma:
This PR updates the DSWx-HLS code after the beta delivery:
- Add error tolerance to function compare_dswx_hls_products().
- Fix bug printing wrong layer name in function compare_dswx_hls_products().
- Remove the creation of directories input_dir, scratch_dir, runconfig, and scratch_dir inside Docker container.
- Fix value assignment of DSWx-HLS products metadata SENSOR_PRODUCT_ID.
- Update comments on DSWx-HLS runconfig schema and default files.
- Update the computation of the joint land cover map from ESA Landcover 100m and ESA Worldcover 10m.
- Add code that masks the DSWx-HLS interpreted layer using the joint land cover map.


gshiroma:
Hi @hfattahi and @collinss-jpl , this PR should be ready for review. 

The changes that would affect the PGE are the following:
- New runconfig field `worldcover_file`. See it [here](https://github.com/opera-adt/PROTEUS/blob/354bb8d891ab7b571f6dc77176cd5a2f84ff685a/src/proteus/schemas/dswx_hls.yaml#L20).
- Instead of using the user `conda:conda` when running the docker image, now you should be able to use `$UID:$(id -g)`.

collinss-jpl:
Hi @gshiroma, just a heads up that I've given the branch a quick look and everything looks good to me so far. I've also been able to build the SAS container without issue. My one question is if you have sample land/world cover ancillary files to use with the sample HLS test datasets, similar to the DEM files that came with the last delivery?

gshiroma:
Thank you so much, @collinss-jpl ! I just sent you an email (using JPL LFT) with a dataset that includes the new land cover maps and can be used to test this PR.

bin/dswx_landcover_mask.py
```diff
@@ -28,14 +30,14 @@ def _get_parser():
                         help='Input HLS product')
 
     parser.add_argument('-c',
-                        '--copernicus-landcover-100m',
+                        '--landcover', '--land-cover',
                         dest='copernicus_landcover_file',
                         required=True,
                         type=str,
                         help='World cover')
```
hfattahi, line 37: would you please add a bit more help here to clarify what is this. Specially we want it to be very clear and distinguished from the next one (--world-cover)

hfattahi (approved):
Thank you @gshiroma. Looks great to me.

collinss-jpl (approved):
Hi @gshiroma, finally had a chance to rerun a successful acceptance test with this build, logs from the test as well as the comparison with the expected results are attached.

[s30_louisiana.log](https://github.com/opera-adt/PROTEUS/files/8702862/s30_louisiana.log)
[s30_louisiana_compare_results.log](https://github.com/opera-adt/PROTEUS/files/8702864/s30_louisiana_compare_results.log)

I only ran into one minor snag with the build: because the Dockerfile no longer creates `scratch_dir`, I got a permission error when PROTEUS tries to create it within `/home/conda` (the default location in the sample `dsxw_hls.yaml`). This is because `/home/conda` is owned by `conda:conda`, whereas we run the container as `$UID:$(id -g)`. The simple workaround was to specify the scratch directory to be `output_dir/scratch_dir` in the runconfig. This works since `output_dir` is mounted into the container with `$UID:$(id -g)` as the owner and group, so everything lines up permission wise. You may want to make this the default for your runconfigs going forward.

Lastly, a couple questions:
- Regarding the sample landcover/worldcover files you provided (thanks again btw), would these files be usable for all DSWx-HLS jobs, or are they somehow specific to the s30_louisiana dataset? If its the latter, could you provide some info on what cropping or translation is necessary to stage the appropriate region from the global masks?
- One of our developers noticed that the `dswx_hls_v0.1_B10_DEM.tif` output product does not include the same set of metadata fields (such as `HLS_DATASET`, `SPACECRAFT_NAME`, etc) as the other output products. Was this intentional? We've been making the assumption that all output bands will have equivalent sets of metadata, so we can pick any of them to extract from, but it seems this is not the case.

gshiroma:
> Hi @gshiroma, finally had a chance to rerun a successful acceptance test with this build, logs from the test as well as the comparison with the expected results are attached.
> 
> [s30_louisiana.log](https://github.com/opera-adt/PROTEUS/files/8702862/s30_louisiana.log) [s30_louisiana_compare_results.log](https://github.com/opera-adt/PROTEUS/files/8702864/s30_louisiana_compare_results.log)
> 
> I only ran into one minor snag with the build: because the Dockerfile no longer creates `scratch_dir`, I got a permission error when PROTEUS tries to create it within `/home/conda` (the default location in the sample `dsxw_hls.yaml`). This is because `/home/conda` is owned by `conda:conda`, whereas we run the container as `$UID:$(id -g)`. The simple workaround was to specify the scratch directory to be `output_dir/scratch_dir` in the runconfig. This works since `output_dir` is mounted into the container with `$UID:$(id -g)` as the owner and group, so everything lines up permission wise. You may want to make this the default for your runconfigs going forward.
> 
> Lastly, a couple questions:
> 
> * Regarding the sample landcover/worldcover files you provided (thanks again btw), would these files be usable for all DSWx-HLS jobs, or are they somehow specific to the s30_louisiana dataset? If its the latter, could you provide some info on what cropping or translation is necessary to stage the appropriate region from the global masks?
> * One of our developers noticed that the `dswx_hls_v0.1_B10_DEM.tif` output product does not include the same set of metadata fields (such as `HLS_DATASET`, `SPACECRAFT_NAME`, etc) as the other output products. Was this intentional? We've been making the assumption that all output bands will have equivalent sets of metadata, so we can pick any of them to extract from, but it seems this is not the case.

Thank you so much for your review, @collinss-jpl ! Thank you also for the log files. I'm glad to see that the DEM comparison went smooth this time.

Thank you for your comment on the `scratch_dir`. If you mounted it the same way as you did for `output_dir`, shouldn't it work? In any case, that's a great idea to default it to `output_dir/scratch_dir`. I'll update it accordingly for this and the next deliveries.

In regards to the land cover maps, the "Copernicus land cover map 100m" is near-global (lon: -180 to 180 deg, lat: -60 to 80 deg) so it should cover any HLS dataset that will be used in production. However, these datasets may change annually. The ESA Worldcover 10m dataset covers a range of 3 deg (lat) x 3 deg (lon). Similarly to the DEM, both datasets could be cropped to avoid unnecessary file movement to the machine that will run the PGE/SAS. One can create a VRT that points to all ESA Worldcover 10m tiles using [GDAL build VRT](https://gdal.org/programs/gdalbuildvrt.html) that can be accessed via CLI or Python API. The SAS Design Document (delivered along beta delivery) provides more information on the land cover dataset as well as how to download these files. Please, let me know if you need more information about staging these datasets.

Good catch on the DEM layer metadata! Some of the layers use a different writer function and the metadata is in fact missing for the DEM. Since this is probably not major, I'd rather fix this in another PR.

Thank you again!

gshiroma:
Hi @hfattahi , it seems that I lost your approval when I updated the argparse description (to address your review). Would you be able to please approve it again? Thank you.

hfattahi (approved):

PR #19, opened 2022-05-02
Switch CircleCI from docker executor to machine executor

rtburns-jpl:
The docker executor does not support docker volume mounting within

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,15 +2,17 @@ version: 2.1
 
 jobs:
   build:
-    docker:
-      - image: docker:stable-git
+    machine:
+      image: ubuntu-2004:current
+    resource_class: medium
     steps:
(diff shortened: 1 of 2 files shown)


rtburns-jpl:
Okay, looks like this has resolved the docker volume mounting issue. Apparently this is a known caveat of the docker-type executors on circleci, and you need to use a machine-type executor to have volume mounting capabilities for docker commands run within. The commands are now working the same as they did when I tested locally. Nice work on the pytests @gshiroma, and sorry they've taken so long to integrate with our CI system!

gshiroma (approved):
This is awesome, @rtburns-jpl ! Thank you so much for finding this and fixing the CI system to run the tests!

hfattahi (approved):
Thank you @rtburns-jpl . Great.

PR #21, opened 2022-05-21
Update the DSWx-HLS SAS for cal/val release

gshiroma:
This PR updates the DSWx-HLS SAS after the mid-May point release:
- Populate DSWx-HLS metadata fields `dem_file`, `landcover_file`, and `worldcover_file`.
- (TODO) Add DSWx-HLS metadata to all DSWx-HLS product layers.

### DIFF
--- a/bin/dswx_landcover_mask.py
+++ b/bin/dswx_landcover_mask.py
@@ -90,8 +90,7 @@ def _get_parser():
 
 def point2epsg(lon, lat):
     """
-     Return EPSG code base on a point
-     latitude/longitude coordinates
+    Return EPSG code based on a point lat/lon coordinates
     """
     if lon >= 180.0:
         lon = lon - 360.0
@@ -114,7 +113,7 @@ def main():
     create_logger(args.log_file)
 
     if args.bbox is None and args.reference_file is None:
-        logger.error(f'ERROR please, provide either a '
+        logger.error('ERROR please, provide either a '
                      'reference file or a bounding box')
         return
 
@@ -185,7 +184,7 @@ def main():
         x_max = None
         for lat in [lat_max, lat_min]:
             for lon in [lon_min, lon_max]:
-                x, y, z = transformation.TransformPoint(lat, lon, 0)
+                x, y, _ = transformation.TransformPoint(lat, lon, 0)
                 if y_min is None or y_min > y:
                     y_min = y
                 if y_max is None or y_max < y:
@@ -206,7 +205,7 @@ def main():
         width = int(np.ceil((x_max - x_min) / dx))
         length = int(np.ceil((y_min - y_max) / dy))
         projection = utm_srs.ExportToProj4()
-        projection = projection.strip() 
+        projection = projection.strip()
 
         print(f'    width: {width}')
         print(f'    length: {length}')
--- a/bin/dswx_landcover_mask.py
+++ b/bin/dswx_landcover_mask.py
@@ -12,7 +12,8 @@
 import argparse
 import logging
 import os
-from osgeo import gdal
+from osgeo import gdal, osr
+import numpy as np
 from proteus.dswx_hls import create_landcover_mask, create_logger
 
 logger = logging.getLogger('dswx_hls_landcover_mask')
@@ -25,14 +26,24 @@ def _get_parser():
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 
     # Inputs
-    parser.add_argument('input_file',
+    parser.add_argument('-g', '--ref',
+                        dest='reference_file',
                         type=str,
-                        help='Input HLS product')
+                        help='Reference file for geographic grid'
+                              ' (e.g. HLS product layer)')
 
-    parser.add_argument('-c',
-                        '--copernicus-landcover-100m',
+    parser.add_argument('-b', '--bbox',
+                        type=float,
+                        nargs=4,
+                        dest='bbox',
+                        metavar=('y_min', 'y_max', 'x_min', 'x_max'),
+                        help='Defines the spatial region in '
+                             'the format south north west east.')
+
+    # other parameters
+    parser.add_argument('-c', '--copernicus-landcover-100m',
                         '--landcover', '--land-cover',
-                        dest='landcover_file',
+                        dest='copernicus_landcover_file',
                         required=True,
                         type=str,
                         help='Input Copernicus Land Cover'
@@ -46,8 +57,7 @@ def _get_parser():
                         help='Input ESA WorldCover 10m')
 
     # Outputs
-    parser.add_argument('-o',
-                        '--output-file',
+    parser.add_argument('-o', '--output-file',
                         dest='output_file',
                         required=True,
                         type=str,
@@ -78,6 +88,24 @@ def _get_parser():
     return parser
 
 
+def point2epsg(lon, lat):
+    """
+     Return EPSG code base on a point
+     latitude/longitude coordinates
+    """
+    if lon >= 180.0:
+        lon = lon - 360.0
+    if lat >= 60.0:
+        return 3413
+    elif lat <= -60.0:
+        return 3031
+    elif lat > 0:
+        return 32601 + int(np.round((lon + 177) / 6.0))
+    elif lat < 0:
+        return 32701 + int(np.round((lon + 177) / 6.0))
+    raise ValueError(
+        'Could not determine projection for {0},{1}'.format(lat, lon))
+
 def main():
     parser = _get_parser()
 
@@ -85,17 +113,105 @@ def main():
 
     create_logger(args.log_file)
 
-    logger.info(f'Input file: {args.input_file}')
-    if not os.path.isfile(args.input_file):
-        logger.error(f'ERROR file not found: {args.input_file}')
+    if args.bbox is None and args.reference_file is None:
+        logger.error(f'ERROR please, provide either a '
+                     'reference file or a bounding box')
         return
-    layer_gdal_dataset = gdal.Open(args.input_file, gdal.GA_ReadOnly)
-    if layer_gdal_dataset is None:
-        logger.error(f'ERROR invalid input HLS file: {args.input_file}')
-    geotransform = layer_gdal_dataset.GetGeoTransform()
-    projection = layer_gdal_dataset.GetProjection()
-    length = layer_gdal_dataset.RasterYSize
-    width = layer_gdal_dataset.RasterXSize
+
+    if args.bbox is not None:
+        flag_is_geographic = True
+        lat_min, lat_max, lon_min, lon_max = args.bbox
+        # make sure lat_max > lat_min (since dy < 0)
+        if lat_max < lat_min:
+            lat_min, lat_max = lat_max, lat_min
+
+    geographic_srs = None
+    if args.reference_file:
+        print(f'Reference file: {args.reference_file}')
+
+        if not os.path.isfile(args.reference_file):
+            logger.error(f'ERROR file not found: {args.reference_file}')
+            return
+        layer_gdal_dataset = gdal.Open(args.reference_file, gdal.GA_ReadOnly)
+        if layer_gdal_dataset is None:
+            logger.error(f'ERROR invalid file: {args.reference_file}')
+
+        # read reference image geolocation
+        geotransform = layer_gdal_dataset.GetGeoTransform()
+        projection = layer_gdal_dataset.GetProjection()
+        length = layer_gdal_dataset.RasterYSize
+        width = layer_gdal_dataset.RasterXSize
+
+        # check if geolocation is in projected coordinates, i.e.,
+        # not geographic
+        projection_srs = osr.SpatialReference(wkt=projection)
+
+        flag_is_geographic = projection_srs.IsGeographic()
+        if flag_is_geographic:
+            geographic_srs = projection_srs
+            print('Reference file is provided in geographic'
+                  ' coordinates.')
+            lat_max = geotransform[3]
+            lat_min = lat_max + geotransform[5] * length
+            lon_min = geotransform[0]
+            lon_max = lon_min + geotransform[1] * width
+        else:
+            print('Reference file is NOT provided in geographic'
+                        ' coordinates')
+
+    if flag_is_geographic:
+        mean_y = (lat_max + lat_min) / 2.0
+        mean_x = (lon_min + lon_max) / 2.0
+        epsg = point2epsg(mean_x, mean_y)
+
+        print('Converting geographic coordinates to UTM:')
+        print(f'    lat_min: {lat_min}, lat_max: {lat_max}')
+        print(f'    lon_min: {lon_min}, lon_max: {lon_max}')
+        print(f'    EPSG code: {epsg}')
+
+        if geographic_srs is None:
+            geographic_srs = osr.SpatialReference()
+            geographic_srs.SetWellKnownGeogCS("WGS84")
+
+        utm_srs = osr.SpatialReference()
+        utm_srs.ImportFromEPSG(epsg)
+
+        # create transformation of coordinates from geographic (lat/lon)
+        # to UTM
+        transformation = osr.CoordinateTransformation(geographic_srs, utm_srs)
+        y_min = None
+        y_max = None
+        x_min = None
+        x_max = None
+        for lat in [lat_max, lat_min]:
+            for lon in [lon_min, lon_max]:
+                x, y, z = transformation.TransformPoint(lat, lon, 0)
+                if y_min is None or y_min > y:
+                    y_min = y
+                if y_max is None or y_max < y:
+                    y_max = y
+                if x_min is None or x_min > x:
+                    x_min = x
+                if x_max is None or x_max < x:
+                    x_max = x
+
+        print(f'    y_min: {y_min}, y_max: {y_max}')
+        print(f'    x_min: {x_min}, x_max: {x_max}')
+
+        # land cover map step is 30m meters
+        dx = 30
+        dy = -30
+        geotransform = [x_min, dx, 0, y_max, 0, dy]
+
+        width = int(np.ceil((x_max - x_min) / dx))
+        length = int(np.ceil((y_min - y_max) / dy))
+        projection = utm_srs.ExportToProj4()
+        projection = projection.strip() 
+
+        print(f'    width: {width}')
+        print(f'    length: {length}')
+        print(f'    geotransform: {geotransform}')
+        print(f'    projection: {projection}')
 
     create_landcover_mask(args.copernicus_landcover_file,
                           args.worldcover_file, args.output_file,
--- a/bin/dswx_landcover_mask.py
+++ b/bin/dswx_landcover_mask.py
@@ -89,9 +89,6 @@ def _get_parser():
 
 
 def point2epsg(lon, lat):
-    """
-    Return EPSG code based on a point lat/lon coordinates
-    """
     if lon >= 180.0:
         lon = lon - 360.0
     if lat >= 60.0:
--- a/bin/dswx_landcover_mask.py
+++ b/bin/dswx_landcover_mask.py
@@ -36,7 +36,7 @@ def _get_parser():
                         type=float,
                         nargs=4,
                         dest='bbox',
-                        metavar=('y_min', 'y_max', 'x_min', 'x_max'),
+                        metavar=('LAT_MIN', 'LAT_MAX', 'LON_MIN', 'LON_MAX'),
                         help='Defines the spatial region in '
                              'the format south north west east.')
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -25,7 +25,7 @@ RUN set -ex \
  && conda update --all --yes \
  && conda install --yes --file /tmp/requirements.txt \
  && conda install --yes --channel conda-forge --file /tmp/requirements_forge.txt \
- && conda clean -tipsy \
+ && conda clean -tip \
  && rm -rf /opt/conda/pkgs \
  && rm /tmp/requirements.txt \
  && rm /tmp/requirements_forge.txt
--- a/src/proteus/core.py
+++ b/src/proteus/core.py
@@ -46,7 +46,7 @@ def save_as_cog(filename, scratch_dir = '.', logger = None,
     del gdal_ds  # close the dataset (Python object and pointers)
     external_overview_file = filename + '.ovr'
     if os.path.isfile(external_overview_file):
-        os.path.remove(external_overview_file)
+        os.remove(external_overview_file)
 
     logger.info('COG step 2: save as COG')
     temp_file = tempfile.NamedTemporaryFile(
@@ -101,7 +101,6 @@ def get_geographic_boundaries_from_mgrs_tile(mgrs_tile_name, verbose=False):
 
     # create UTM spatial reference
     utm_coordinate_system = osr.SpatialReference()
-    utm_coordinate_system.SetWellKnownGeogCS("WGS84")
     utm_coordinate_system.SetUTM(utm_zone, is_northern)
 
     # create geographic (lat/lon) spatial reference
--- a/src/proteus/core.py
+++ b/src/proteus/core.py
@@ -53,13 +53,12 @@ def save_as_cog(filename, scratch_dir = '.', logger = None,
                     dir=scratch_dir, suffix='.tif').name
 
     tile_size = 256
-    ovr_tile_size = tile_size
-    gdal_translate_options = [
-        'TILED=YES',
-        f'BLOCKXSIZE={tile_size}',
-        f'BLOCKYSIZE={tile_size}',
-        f'GDAL_TIFF_OVR_BLOCKSIZE={ovr_tile_size}'
-        'COPY_SRC_OVERVIEWS=YES'] 
+    # ovr_tile_size = tile_size
+    gdal_translate_options = ['TILED=YES',
+                              f'BLOCKXSIZE={tile_size}',
+                              f'BLOCKYSIZE={tile_size}',
+                              # f'GDAL_TIFF_OVR_BLOCKSIZE={ovr_tile_size}'
+                              'COPY_SRC_OVERVIEWS=YES'] 
 
     if flag_compress:
         gdal_translate_options += ['COMPRESS=DEFLATE']
@@ -67,7 +66,7 @@ def save_as_cog(filename, scratch_dir = '.', logger = None,
     if is_integer:
         gdal_translate_options += ['PREDICTOR=2']
     else:
-        gdal_translate_options = ['PREDICTOR=3']
+        gdal_translate_options += ['PREDICTOR=3']
 
     gdal.Translate(temp_file, filename,
                    creationOptions=gdal_translate_options)
--- a/src/proteus/core.py
+++ b/src/proteus/core.py
@@ -40,7 +40,7 @@ def save_as_cog(filename, scratch_dir = '.', logger = None,
     else:
         resamp_algorithm = 'CUBICSPLINE'
 
-    gdal_ds.BuildOverviews('CUBICSPLINE', overviews_list,
+    gdal_ds.BuildOverviews(resamp_algorithm, overviews_list,
                            gdal.TermProgress_nocb)
 
     del gdal_ds  # close the dataset (Python object and pointers)
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -1478,7 +1478,7 @@ def _get_binary_water_ctable():
     # Water
     binary_water_ctable.SetColorEntry(1, (0, 0, 255))
     # Gray - QA masked
-    binary_water_ctable.SetColorEntry(2, (127, 127, 127))
+    binary_water_ctable.SetColorEntry(9, (127, 127, 127))
     # Black - Fill value
     binary_water_ctable.SetColorEntry(255, (0, 0, 0, 255))
     return binary_water_ctable
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -18,7 +18,7 @@ landcover_mask_type = 'standard'
 
 COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE = 1e-6
 
-DEM_MARGIN_IN_PIXELS = 5
+DEM_MARGIN_IN_PIXELS = 10
 
 logger = logging.getLogger('dswx_hls')
 
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -56,6 +56,44 @@ s30_v2_band_dict = {'blue': 'B02',
                     'qa': 'Fmask'}
 
 interpreted_dswx_band_dict = {
+    0b00000 : 0,  # (not water)
+    0b00001 : 0,
+    0b00010 : 0,
+    0b00100 : 0,
+    0b01000 : 0,
+    0b01111 : 1,  # (open water)
+    0b10111 : 1,
+    0b11011 : 1,
+    0b11101 : 1,
+    0b11110 : 1,
+    0b11111 : 1,
+    0b00111 : 1,
+    0b01011 : 1,
+    0b01101 : 1,
+    0b01110 : 1,
+    0b10011 : 1,
+    0b10101 : 1,
+    0b10110 : 1,
+    0b11001 : 1,
+    0b11010 : 1,
+    0b11100 : 1,
+    0b11000 : 2,  # (partial surface water)
+    0b00011 : 2,
+    0b00101 : 2,
+    0b00110 : 2,
+    0b01001 : 2,
+    0b01010 : 2,
+    0b01100 : 2,
+    0b10000 : 2,
+    0b10001 : 2,
+    0b10010 : 2,
+    0b10100 : 2
+    }
+
+'''
+Original mapping (DSWe):
+
+interpreted_dswx_band_dict = {
     0b00000 : 0,  # (Not Water)
     0b00001 : 0,
     0b00010 : 0,
@@ -88,6 +126,7 @@ interpreted_dswx_band_dict = {
     0b10001 : 4,
     0b10010 : 4,
     0b10100 : 4}
+'''
 
 band_description_dict = {
     'WTR': 'Water classification (WTR)',
@@ -852,12 +891,24 @@ def _get_interpreted_dswx_ctable():
     dswx_ctable.SetColorEntry(0, (255, 255, 255))
     # Blue - Water (high confidence)
     dswx_ctable.SetColorEntry(1, (0, 0, 255))
+    # Green - Low confidence water or wetland
+    dswx_ctable.SetColorEntry(2, (0, 255, 0))
+    # Light blue - Reserved
+    dswx_ctable.SetColorEntry(3, (0, 127, 255))
+    # Dark green - Reserved
+    dswx_ctable.SetColorEntry(4, (0, 127, 0))
+
+    '''
+    # Blue - Water (high confidence)
+    dswx_ctable.SetColorEntry(1, (0, 0, 255))
     # Light blue - Water (moderate conf.)
     dswx_ctable.SetColorEntry(2, (0, 127, 255))
     # Dark green - Potential wetland
     dswx_ctable.SetColorEntry(3, (0, 127, 0))
     # Green - Low confidence water or wetland
     dswx_ctable.SetColorEntry(4, (0, 255, 0))
+    '''
+
     # Cyan - Reserved
     dswx_ctable.SetColorEntry(5, (0, 255, 255))
     # Light green - Reserved
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -2463,6 +2463,7 @@ def generate_dswx_layers(input_list,
            _save_array(dem, output_dem_layer,
                        dswx_metadata_dict, geotransform, projection,
                        description=band_description_dict['DEM'],
+                       output_dtype = gdal.GDT_Float32,
                        scratch_dir=scratch_dir,
                        output_files_list=build_vrt_list)
         if not output_file:
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -838,14 +838,15 @@ def _apply_landcover_and_shadow_masks(interpreted_layer, nir,
     # apply shadow mask - shadows are set to 0 (not water)
     if shadow_layer is not None and landcover_mask is None:
         logger.info('applying shadow mask:')
-        to_mask_ind = np.where(shadow_layer == 0 &
-            ((interpreted_layer == 1) | (interpreted_layer == 2)))
+        to_mask_ind = np.where((shadow_layer == 0) &
+            ((interpreted_layer >= 1) | (interpreted_layer <= 4)))
         landcover_shadow_masked_dswx[to_mask_ind] = 0
 
     elif shadow_layer is not None:
+        logger.info('applying shadow mask (with landcover):')
         to_mask_ind = np.where((shadow_layer == 0) &
             (~_is_landcover_class_water_or_wetland(landcover_mask)) &
-            ((interpreted_layer == 1) | (interpreted_layer == 2)))
+            ((interpreted_layer >= 1) & (interpreted_layer <= 4)))
         landcover_shadow_masked_dswx[to_mask_ind] = 0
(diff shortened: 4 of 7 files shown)


collinss-jpl:
Hi @gshiroma, I noticed another discrepancy in the DSWx-HLS metadata regarding the `SENSING_TIME` field. It seems like for L30 datasets the field is set with both a beginning and end time, delimited by semi-colon:

```python
>>> from osgeo import gdal
>>> ds = gdal.Open(".../output_dir/OPERA_L3_DSWx_HLS_LANDSAT-8_T22VEQ_20210905T143156_v2.0_001_B01_WTR.tif")
>>> ds.GetMetadata()['SENSING_TIME']
'2021-09-05T14:31:56.9300799Z; 2021-09-05T14:32:20.8126470Z'
```
This is fine, as we want a begin/end time to put into the output ISO XML.

However, for an S30 dataset, only a single time is available:
```python
>>> ds = gdal.Open(".../output_dir/OPERA_L3_DSWx_HLS_SENTINEL-2A_T15SXR_20210907T163901_v2.0_001_B01_WTR.tif")
>>> ds.GetMetadata()['SENSING_TIME']
'2021-09-07T16:54:47.751044Z'
```
Is this the expected behavior for S30 datasets, or possibly a bug?

gshiroma:
> Hi @gshiroma, I noticed another discrepancy in the DSWx-HLS metadata regarding the `SENSING_TIME` field. It seems like for L30 datasets the field is set with both a beginning and end time, delimited by semi-colon:
> 
> ```python
> >>> from osgeo import gdal
> >>> ds = gdal.Open(".../output_dir/OPERA_L3_DSWx_HLS_LANDSAT-8_T22VEQ_20210905T143156_v2.0_001_B01_WTR.tif")
> >>> ds.GetMetadata()['SENSING_TIME']
> '2021-09-05T14:31:56.9300799Z; 2021-09-05T14:32:20.8126470Z'
> ```
> 
> This is fine, as we want a begin/end time to put into the output ISO XML.
> 
> However, for an S30 dataset, only a single time is available:
> 
> ```python
> >>> ds = gdal.Open(".../output_dir/OPERA_L3_DSWx_HLS_SENTINEL-2A_T15SXR_20210907T163901_v2.0_001_B01_WTR.tif")
> >>> ds.GetMetadata()['SENSING_TIME']
> '2021-09-07T16:54:47.751044Z'
> ```
> 
> Is this the expected behavior for S30 datasets, or possibly a bug?

Hi @collinss-jpl , thank you so much again for reporting this. Good point, the metadata field `SENSING_TIME` is imported from the input HLS product and, as you also pointed out, they differ between L30 and S30 datasets. I'll make sure to make them consistent. I'll probably keep the original value for S30 datasets and I'll take the average value for L30 datasets. Thanks again.

gshiroma:
> Hi @gshiroma, I noticed another discrepancy in the DSWx-HLS metadata regarding the `SENSING_TIME` field. It seems like for L30 datasets the field is set with both a beginning and end time, delimited by semi-colon:
> 
> ```python
> >>> from osgeo import gdal
> >>> ds = gdal.Open(".../output_dir/OPERA_L3_DSWx_HLS_LANDSAT-8_T22VEQ_20210905T143156_v2.0_001_B01_WTR.tif")
> >>> ds.GetMetadata()['SENSING_TIME']
> '2021-09-05T14:31:56.9300799Z; 2021-09-05T14:32:20.8126470Z'
> ```
> 
> This is fine, as we want a begin/end time to put into the output ISO XML.
> 
> However, for an S30 dataset, only a single time is available:
> 
> ```python
> >>> ds = gdal.Open(".../output_dir/OPERA_L3_DSWx_HLS_SENTINEL-2A_T15SXR_20210907T163901_v2.0_001_B01_WTR.tif")
> >>> ds.GetMetadata()['SENSING_TIME']
> '2021-09-07T16:54:47.751044Z'
> ```
> 
> Is this the expected behavior for S30 datasets, or possibly a bug?

Just created an issue #23 to address this comment. Thanks.

gshiroma:
> Hi @gshiroma, I noticed another discrepancy in the DSWx-HLS metadata regarding the `SENSING_TIME` field. It seems like for L30 datasets the field is set with both a beginning and end time, delimited by semi-colon:
> 
> ```python
> >>> from osgeo import gdal
> >>> ds = gdal.Open(".../output_dir/OPERA_L3_DSWx_HLS_LANDSAT-8_T22VEQ_20210905T143156_v2.0_001_B01_WTR.tif")
> >>> ds.GetMetadata()['SENSING_TIME']
> '2021-09-05T14:31:56.9300799Z; 2021-09-05T14:32:20.8126470Z'
> ```
> 
> This is fine, as we want a begin/end time to put into the output ISO XML.
> 
> However, for an S30 dataset, only a single time is available:
> 
> ```python
> >>> ds = gdal.Open(".../output_dir/OPERA_L3_DSWx_HLS_SENTINEL-2A_T15SXR_20210907T163901_v2.0_001_B01_WTR.tif")
> >>> ds.GetMetadata()['SENSING_TIME']
> '2021-09-07T16:54:47.751044Z'
> ```
> 
> Is this the expected behavior for S30 datasets, or possibly a bug?

Hi @collinss-jpl , commit [602fc35](https://github.com/opera-adt/PROTEUS/pull/21/commits/602fc35a5edeeba9d0fa0832546fc864a2599acb) fixes this issue. Thank you again for reporting it.

gshiroma:
Hi @hfattahi, @collinss-jpl , and @nemo794 . This PR is ready to be reviewed. Sorry for the delay. I was making sure we were addressing all known code issues before this cal/val delivery.

bin/dswx_landcover_mask.py
```diff
@@ -25,14 +26,24 @@ def _get_parser():
...
+                              ' (e.g. HLS product layer)')
+
+    parser.add_argument('-b', '--bbox',
+                        type=float,
+                        nargs=4,
+                        dest='bbox',
+                        metavar=('LAT_MIN', 'LAT_MAX', 'LON_MIN', 'LON_MAX'),
+                        help='Defines the spatial region in '
+                             'the format south north west east.')
+
```
nemo794, line 42:
```suggestion
                        metavar=('LON_MIN', 'LAT_MIN', 'LON_MAX', 'LAT_MAX'),
                        help='Defines the spatial bounding box region in '
                             'the NASA STAC-CMR format of LL (lower left) '
                             'and UR (upper right) coordinates, respectively.')

```

bin/dswx_landcover_mask.py
```diff
@@ -78,24 +88,126 @@ def _get_parser():
...
-    if layer_gdal_dataset is None:
-        logger.error(f'ERROR invalid input HLS file: {args.input_file}')
-    geotransform = layer_gdal_dataset.GetGeoTransform()
-    projection = layer_gdal_dataset.GetProjection()
-    length = layer_gdal_dataset.RasterYSize
-    width = layer_gdal_dataset.RasterXSize
+
+    if args.bbox is not None:
+        flag_is_geographic = True
+        lat_min, lat_max, lon_min, lon_max = args.bbox
```
nemo794, line 119:
```suggestion
        lon_min, lat_min, lon_max, lat_min = args.bbox
```

bin/dswx_landcover_mask.py
```diff
@@ -78,24 +88,126 @@ def _get_parser():
...
+    geographic_srs = None
+    if args.reference_file:
+        print(f'Reference file: {args.reference_file}')
+
+        if not os.path.isfile(args.reference_file):
+            logger.error(f'ERROR file not found: {args.reference_file}')
+            return
+        layer_gdal_dataset = gdal.Open(args.reference_file, gdal.GA_ReadOnly)
+        if layer_gdal_dataset is None:
+            logger.error(f'ERROR invalid file: {args.reference_file}')
```
nemo794, line 133: If layer_gdal_dataset is None, then that error is logged to file, but the rest of the code is allowed to proceed. (But will quickly error-out, such as in line 136). Is there a missing `return` statement, similar to line 130? If not, what should the desired behavior be?

bin/dswx_landcover_mask.py
```diff
@@ -78,24 +88,126 @@ def _get_parser():
...
+                     'reference file or a bounding box')
         return
-    layer_gdal_dataset = gdal.Open(args.input_file, gdal.GA_ReadOnly)
-    if layer_gdal_dataset is None:
-        logger.error(f'ERROR invalid input HLS file: {args.input_file}')
-    geotransform = layer_gdal_dataset.GetGeoTransform()
-    projection = layer_gdal_dataset.GetProjection()
-    length = layer_gdal_dataset.RasterYSize
-    width = layer_gdal_dataset.RasterXSize
+
```
nemo794, line 116:
```suggestion

flag_is_geographic = False
```

bin/dswx_landcover_mask.py
```diff
@@ -78,24 +88,126 @@ def _get_parser():
...
+    if args.bbox is not None:
+        flag_is_geographic = True
+        lat_min, lat_max, lon_min, lon_max = args.bbox
+        # make sure lat_max > lat_min (since dy < 0)
+        if lat_max < lat_min:
+            lat_min, lat_max = lat_max, lat_min
+
+    geographic_srs = None
```
nemo794, line 124:
```suggestion
    geographic_srs = None
    flag_is_geographic = False
```

docker/Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN set -ex \
  && conda update --all --yes \
  && conda install --yes --file /tmp/requirements.txt \
  && conda install --yes --channel conda-forge --file /tmp/requirements_forge.txt \
- && conda clean -tipsy \
+ && conda clean -tip \
```
nemo794, line 28: Optional: can make this `-tipy` if you'd like to keep the `--yes` behavior.

jshimada47:
@gshiroma When I ran build_docker_image.sh on aurora this afternoon, tests/test_dswx_hls_workflow.py failed with a 'Temporary failure in name resolution' error:
```
>           raise ConnectionError(e, request=request)
E           requests.exceptions.ConnectionError: HTTPSConnectionPool(host='zenodo.org', port=443): Max retries exceeded with url: /record/
6819971/files/s30_louisiana_mississippi.tar.gz (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f24a7600190>: Fail
ed to establish a new connection: [Errno -3] Temporary failure in name resolution'))

/opt/conda/lib/python3.9/site-packages/requests/adapters.py:565: ConnectionError
------------------------------------------------------------- Captured stdout call --------------------------------------------------------------

Test dataset s30_louisiana_mississippi not found. Downloading file https://zenodo.org/record/6819971/files/s30_louisiana_mississippi.tar.gz.
============================================================ short test summary info ============================================================

FAILED tests/test_dswx_hls_workflow.py::test_workflow - requests.exceptions.ConnectionError: HTTPSConnectionPool(host='zenodo.org', port=443):...

========================================================== 1 failed, 1 passed in 5.59s ==========================
```
There were several of these errors; the above is the last one.  I first verified that the zenodo.org URL was OK.  Then I remembered that a similar error had occurred when I built the CSLC_S1 docker image on aurora last month.  The problem then and today was that 'docker run' was unable to resolve DNS hostnames.  There are several possible solutions, but a simple fix is to add the option '--network host' to the 'docker run' command in build_docker_image.sh, such as in
```
docker run --rm -u "$(id -u):$(id -g)" -v "$PWD:/mnt" -w /mnt -it --network host "${IMAGE}:$t" pytest /mnt/tests/
```
This option tells docker to have the host (aurora) perform DNS lookups.  After making the above change, I was able to build the docker image and successfully test it with the Greenland and Louisiana data sets.

Some of the other possible solutions require root privileges, e.g., rebooting the system or modifying various system files.  Aurora itself hasn't been rebooted for several months.  However, rebooting it or modifying its system files isn't a practical option for us.  It's also not clear why docker sometimes loses the ability to resolve hostnames.

src/proteus/core.py
```diff
@@ -40,34 +40,33 @@ def save_as_cog(filename, scratch_dir = '.', logger = None,
...
 
     tile_size = 256
-    ovr_tile_size = tile_size
-    gdal_translate_options = [
-        'TILED=YES',
-        f'BLOCKXSIZE={tile_size}',
-        f'BLOCKYSIZE={tile_size}',
-        f'GDAL_TIFF_OVR_BLOCKSIZE={ovr_tile_size}'
-        'COPY_SRC_OVERVIEWS=YES'] 
+    # ovr_tile_size = tile_size
```
nemo794, line 56:
```suggestion
```

nemo794, line 56: Could you please add a comment explaining why 256 was chosen? Thank you!

src/proteus/core.py
```diff
@@ -40,34 +40,33 @@ def save_as_cog(filename, scratch_dir = '.', logger = None,
...
+    gdal_translate_options = ['TILED=YES',
+                              f'BLOCKXSIZE={tile_size}',
+                              f'BLOCKYSIZE={tile_size}',
+                              # f'GDAL_TIFF_OVR_BLOCKSIZE={ovr_tile_size}'
```
nemo794, line 60:
```suggestion
```

tests/test_dswx_hls_workflow.py
```diff
@@ -91,4 +91,4 @@ def test_workflow():
 
         assert compare_dswx_hls_products(ref_file, output_file)
 
-    assert True
\ No newline at end of file
+    assert True
```
nemo794, line 94: Please add newline at end of file.

nemo794, line 94: Out of curiosity, what is the reason for an `assert True` at the end? This should always pass; is it something to do with the test environment? Thanks!

nemo794 (commented):
Next batch of suggestions, all very small.

gshiroma:
> Next batch of suggestions, all very small.

Thank you so much for all the suggestions, @nemo794 ! I don't see any critical change for now. I'll leave all comments here so we can address each of them in the next PR. Once you are done reviewing this PR, if you don't see anything critical, I'd ask you to please approve this PR so we can have it merged for this cal/val delivery.

src/proteus/dswx_hls.py
```diff
@@ -545,7 +583,8 @@ def _update_landcover_array(conglomerate_array, agg_sum, threshold,
 def create_landcover_mask(copernicus_landcover_file,
                           worldcover_file, output_file, scratch_dir,
                           mask_type, geotransform, projection, length, width,
-                          dswx_metadata_dict = {}, output_files_list = None):
+                          dswx_metadata_dict = {}, output_files_list = None,
```
nemo794, line 586: For Python, it is usually best to not use mutable objects (such as dicts or lists) as defaults, because the object will persist between calls to the function. Here's a short example to illustrate: https://dollardhingra.com/blog/python-mutable-default-arguments/ 

If the goal is to have the dictionary persist, could that intended behavior please be documented in the doc string? Otherwise, perhaps set `dswx_metadata_dict=None` and then use a `if dswx_metadata_dict=None: dswx_metadata_dict = {}` condition in the function's body.

src/proteus/dswx_hls.py
```diff
@@ -18,6 +19,8 @@
 
 COMPARE_DSWX_HLS_PRODUCTS_ERROR_TOLERANCE = 1e-6
 
+DEM_MARGIN_IN_PIXELS = 10
```
nemo794, line 22: Out of curiosity, was there a reason that 10 was chosen? Could be helpful to document this, for when people ask in the future.

src/proteus/dswx_hls.py
```diff
@@ -1055,11 +1104,42 @@ def _get_binary_water_layer(interpreted_water_layer):
...
+            Confidence layer
+    """
+    if flag_collapse_wtr_classes:
+        confidence_layer_classes = wtr_confidence_dict
+    else:
+        confidence_layer_classes = wtr_confidence_non_collapsed_dict
+    confidence_layer = np.full_like(interpreted_layer, 255)
+    for original_value, new_value in confidence_layer_classes.items():
+        ind = np.where(interpreted_layer == original_value)
+        confidence_layer[ind] = new_value
```
nemo794, line 1139:
```suggestion
        confidence_layer[interpreted_layer == original_value] = new_value
```

src/proteus/dswx_hls.py
```diff
@@ -1116,38 +1196,35 @@ def _compute_diagnostic_tests(blue, green, red, nir, swir1, swir2,
...
-            if (mndwi[i, j] > hls_thresholds.pswt_2_mndwi and
-                    blue[i, j] < hls_thresholds.pswt_2_blue and
-                    swir1[i, j] < hls_thresholds.pswt_2_swir1 and
-                    swir2[i, j] < hls_thresholds.pswt_2_swir2 and
-                    nir[i, j] < hls_thresholds.pswt_2_nir):
-                diagnostic_layer[i, j] += 16
-
+    # Surface water tests (see [1, 2])
+
+    # Test 1 (open water test, more conservative)
```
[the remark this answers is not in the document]
nemo794: It might reduce readability, but would improve performance to skip the np.where(..) step and simply put the existing conditionals inside the arrays' brackets. (see the comment on line 138 for an example.)

nemo794 (commented):
So much great work here! I've added some additional thoughts, primarily to help with code clarity and maintainability, but a couple of questions as well. (Also, some of the coding-style things might only be noted once, but should be applied broadly). Please take a look a see what needs to be addressed before the release, and please let me know if you have any questions. Thanks!

collinss-jpl (commented):
Code changes look good to me from a PGE perspective, and I was able to create the image from this branch and run a PGE integration test without issues, so this looks good-to-go from my end.

collinss-jpl (approved):

gshiroma:
Thank you so much, @nemo794 ! I looked at all the comments. You caught so many things! All comments are very helpful, and I'll make sure to address all of them in the next release. 

I didn't find anything critical for this cal/val delivery. Since the current code was tested by @jshimada47 , @collinss-jpl , and me, I'd prefer to update the code now. 

Would it be ok to defer these changes to the next delivery? I can create an issue to make sure we'll address these comments. Is that ok, @hfattahi ?

If you see anything critical please let me know.

gshiroma:
> Code changes look good to me from a PGE perspective, and I was able to create the image from this branch and run a PGE integration test without issues, so this looks good-to-go from my end.

Thank you so much, @collinss-jpl . I appreciate you taking the time to test this code for PGE integration.

nemo794 (approved):
This PR is approved so that it can be merged for the cal-val point release. As discussed, there is nothing critical for this release, and so the comments will need to be opened in a new issue and be addressed in the next release. Thanks again, Gustavo!

gshiroma:
> This PR is approved so that it can be merged for the cal-val point release. As discussed, there is nothing critical for this release, and so the comments will need to be opened in a new issue and be addressed in the next release. Thanks again, Gustavo!

Thank you so much, @nemo794, for your in-depth review! I created issue #24 to make sure that we'll address all your comments.

@hfattahi , do you think we can merge this PR?

collinss-jpl:
Hi again @gshiroma, sorry to say but I have another finding that needs to be addressed before this branch is merged. It looks like the `PRODUCT_VERSION` constant in the code has not been updated to match the version tags getting used with github: see https://github.com/nasa/PROTEUS/blob/main/src/proteus/dswx_hls.py#L2141 where the version is still set to 0.1, whereas the version used for the "mid_may" release is 0.3.

Since the PGE extracts this version from the tiff metadata for the output filenames, its very important that these versions are kept in sync. The PST team in particular is concerned with leveraging this version number so they can track any changes made to HLS parameters, etc...

gshiroma:
> Hi again @gshiroma, sorry to say but I have another finding that needs to be addressed before this branch is merged. It looks like the `PRODUCT_VERSION` constant in the code has not been updated to match the version tags getting used with github: see https://github.com/nasa/PROTEUS/blob/main/src/proteus/dswx_hls.py#L2141 where the version is still set to 0.1, whereas the version used for the "mid_may" release is 0.3, and I'm guessing will become 0.4 for this CalVal release.
> 
> Since the PGE extracts this version from the tiff metadata for the output filenames, its very important that these versions are kept in sync. The PST team in particular is concerned with leveraging this version number so they can track any changes made to HLS parameters, etc...

Hi @collinss-jpl , thank you for your comment. The `PRODUCT_VERSION` is not related to the PROTEUS tag. We are still reviewing how we are going to number the `PRODUCT_VERSION`.

collinss-jpl:
> > Hi again @gshiroma, sorry to say but I have another finding that needs to be addressed before this branch is merged. It looks like the `PRODUCT_VERSION` constant in the code has not been updated to match the version tags getting used with github: see https://github.com/nasa/PROTEUS/blob/main/src/proteus/dswx_hls.py#L2141 where the version is still set to 0.1, whereas the version used for the "mid_may" release is 0.3, and I'm guessing will become 0.4 for this CalVal release.
> > Since the PGE extracts this version from the tiff metadata for the output filenames, its very important that these versions are kept in sync. The PST team in particular is concerned with leveraging this version number so they can track any changes made to HLS parameters, etc...
> 
> Hi @collinss-jpl , thank you for your comment. The `PRODUCT_VERSION` is not related to the PROTEUS tag. We are still reviewing how we are going to number the `PRODUCT_VERSION`.

@gshiroma I see, here's a little more background on a potential problem raised by PST in a tag-up we just had. They want to make sure they have a way to easily determine which output products are associated with the default HLS thresholds defined [here](https://github.com/nasa/PROTEUS/blob/main/src/proteus/defaults/dswx_hls.yaml#L61). If, hypothetically, ADT decides that these defaults are to change, could we expect that `PRODUCT_VERSION` would be incremented to indicate as such?

gshiroma:
> > > Hi again @gshiroma, sorry to say but I have another finding that needs to be addressed before this branch is merged. It looks like the `PRODUCT_VERSION` constant in the code has not been updated to match the version tags getting used with github: see https://github.com/nasa/PROTEUS/blob/main/src/proteus/dswx_hls.py#L2141 where the version is still set to 0.1, whereas the version used for the "mid_may" release is 0.3, and I'm guessing will become 0.4 for this CalVal release.
> > > Since the PGE extracts this version from the tiff metadata for the output filenames, its very important that these versions are kept in sync. The PST team in particular is concerned with leveraging this version number so they can track any changes made to HLS parameters, etc...
> > 
> > 
> > Hi @collinss-jpl , thank you for your comment. The `PRODUCT_VERSION` is not related to the PROTEUS tag. We are still reviewing how we are going to number the `PRODUCT_VERSION`.
> 
> @gshiroma I see, here's a little more background on a potential problem raised by PST in a tag-up we just had. They want to make sure they have a way to easily determine which output products are associated with the default HLS thresholds defined [here](https://github.com/nasa/PROTEUS/blob/main/src/proteus/defaults/dswx_hls.yaml#L61). If, hypothetically, ADT decides that these defaults are to change, could we expect that `PRODUCT_VERSION` would be incremented to indicate as such?

HI @collinss-jpl , yes, this is still under discussion. There could be changes to HLS thresholds, the product itself (e.g., metadata fields/layers/structure), and the inputs (DEM, land cover, etc.). We are considering making this field a parameter of the runconfig, so the PGE would set the product version. But we haven't come to any decision now. Do you have any thoughts/suggestions?

collinss-jpl:
> > > > Hi again @gshiroma, sorry to say but I have another finding that needs to be addressed before this branch is merged. It looks like the `PRODUCT_VERSION` constant in the code has not been updated to match the version tags getting used with github: see https://github.com/nasa/PROTEUS/blob/main/src/proteus/dswx_hls.py#L2141 where the version is still set to 0.1, whereas the version used for the "mid_may" release is 0.3, and I'm guessing will become 0.4 for this CalVal release.
> > > > Since the PGE extracts this version from the tiff metadata for the output filenames, its very important that these versions are kept in sync. The PST team in particular is concerned with leveraging this version number so they can track any changes made to HLS parameters, etc...
> > > 
> > > 
> > > Hi @collinss-jpl , thank you for your comment. The `PRODUCT_VERSION` is not related to the PROTEUS tag. We are still reviewing how we are going to number the `PRODUCT_VERSION`.
> > 
> > 
> > @gshiroma I see, here's a little more background on a potential problem raised by PST in a tag-up we just had. They want to make sure they have a way to easily determine which output products are associated with the default HLS thresholds defined [here](https://github.com/nasa/PROTEUS/blob/main/src/proteus/defaults/dswx_hls.yaml#L61). If, hypothetically, ADT decides that these defaults are to change, could we expect that `PRODUCT_VERSION` would be incremented to indicate as such?
> 
> HI @collinss-jpl , yes, this is still under discussion. There could be changes to HLS thresholds, the product itself (e.g., metadata fields/layers/structure), and the inputs (DEM, land cover, etc.). We are considering making this field a parameter of the runconfig, so the PGE would set the product version. But we haven't come to any decision yet. Do you have any thoughts/suggestions?

Ultimately I think it should be up to ADT when and why the version is incremented, so I think specifying it in the RunConfig is not the way to go. 

The main goal is to establish the traceability such that the product version in the output filenames clearly indicates what revision/delivery/configuration of the SAS was used. That was why I thought it should be connected with the release tag version. As long as that traceability is maintained or documented somewhere I think we'll be able to keep the PST team satisfied.

gshiroma:
> > > > > Hi again @gshiroma, sorry to say but I have another finding that needs to be addressed before this branch is merged. It looks like the `PRODUCT_VERSION` constant in the code has not been updated to match the version tags getting used with github: see https://github.com/nasa/PROTEUS/blob/main/src/proteus/dswx_hls.py#L2141 where the version is still set to 0.1, whereas the version used for the "mid_may" release is 0.3, and I'm guessing will become 0.4 for this CalVal release.
> > > > > Since the PGE extracts this version from the tiff metadata for the output filenames, its very important that these versions are kept in sync. The PST team in particular is concerned with leveraging this version number so they can track any changes made to HLS parameters, etc...
> > > > 
> > > > 
> > > > Hi @collinss-jpl , thank you for your comment. The `PRODUCT_VERSION` is not related to the PROTEUS tag. We are still reviewing how we are going to number the `PRODUCT_VERSION`.
> > > 
> > > 
> > > @gshiroma I see, here's a little more background on a potential problem raised by PST in a tag-up we just had. They want to make sure they have a way to easily determine which output products are associated with the default HLS thresholds defined [here](https://github.com/nasa/PROTEUS/blob/main/src/proteus/defaults/dswx_hls.yaml#L61). If, hypothetically, ADT decides that these defaults are to change, could we expect that `PRODUCT_VERSION` would be incremented to indicate as such?
> > 
> > 
> > HI @collinss-jpl , yes, this is still under discussion. There could be changes to HLS thresholds, the product itself (e.g., metadata fields/layers/structure), and the inputs (DEM, land cover, etc.). We are considering making this field a parameter of the runconfig, so the PGE would set the product version. But we haven't come to any decision yet. Do you have any thoughts/suggestions?
> 
> Ultimately I think it should be up to ADT when and why the version is incremented, so I think specifying it in the RunConfig is not the way to go.
> 
> The main goal is to establish the traceability such that the product version in the output filenames clearly indicates what revision/delivery/configuration of the SAS was used. That was why I thought it should be connected with the release tag version. As long as that traceability is maintained or documented somewhere I think we'll be able to keep the PST team satisfied.

Yes, that's a good point, @collinss-jpl . We are just not sure how to describe changes in the product because of updates in the PCM/PGE (including changes in the runconfig parameters populated by the PGE), or updates in any of the inputs, i.e., HLS product with a different version (e.g., v2.1), or updates to the DEM or landcover, for example. Another option is to divide the product version in two parts, one populated by the DSWx-HLS SAS and another by the PGE through the runconfig. Something like `v.1.0.1`, where `1.0` could come from DSWx-HLS SAS and `1` from the PGE. 

Any thoughts @hfattahi ?

hfattahi:
> > > > > > Hi again @gshiroma, sorry to say but I have another finding that needs to be addressed before this branch is merged. It looks like the `PRODUCT_VERSION` constant in the code has not been updated to match the version tags getting used with github: see https://github.com/nasa/PROTEUS/blob/main/src/proteus/dswx_hls.py#L2141 where the version is still set to 0.1, whereas the version used for the "mid_may" release is 0.3, and I'm guessing will become 0.4 for this CalVal release.
> > > > > > Since the PGE extracts this version from the tiff metadata for the output filenames, its very important that these versions are kept in sync. The PST team in particular is concerned with leveraging this version number so they can track any changes made to HLS parameters, etc...
> > > > > 
> > > > > 
> > > > > Hi @collinss-jpl , thank you for your comment. The `PRODUCT_VERSION` is not related to the PROTEUS tag. We are still reviewing how we are going to number the `PRODUCT_VERSION`.
> > > > 
> > > > 
> > > > @gshiroma I see, here's a little more background on a potential problem raised by PST in a tag-up we just had. They want to make sure they have a way to easily determine which output products are associated with the default HLS thresholds defined [here](https://github.com/nasa/PROTEUS/blob/main/src/proteus/defaults/dswx_hls.yaml#L61). If, hypothetically, ADT decides that these defaults are to change, could we expect that `PRODUCT_VERSION` would be incremented to indicate as such?
> > > 
> > > 
> > > HI @collinss-jpl , yes, this is still under discussion. There could be changes to HLS thresholds, the product itself (e.g., metadata fields/layers/structure), and the inputs (DEM, land cover, etc.). We are considering making this field a parameter of the runconfig, so the PGE would set the product version. But we haven't come to any decision yet. Do you have any thoughts/suggestions?
> > 
> > 
> > Ultimately I think it should be up to ADT when and why the version is incremented, so I think specifying it in the RunConfig is not the way to go.
> > The main goal is to establish the traceability such that the product version in the output filenames clearly indicates what revision/delivery/configuration of the SAS was used. That was why I thought it should be connected with the release tag version. As long as that traceability is maintained or documented somewhere I think we'll be able to keep the PST team satisfied.
> 
> Yes, that's a good point, @collinss-jpl . We are just not sure how to describe changes in the product because of updates in the PCM/PGE (including changes in the runconfig parameters populated by the PGE), or updates in any of the inputs, i.e., HLS product with a different version (e.g., v2.1), or updates to the DEM or landcover, for example. Another option is to divide the product version in two parts, one populated by the DSWx-HLS SAS and another by the PGE through the runconfig. Something like `v.1.0.1`, where `1.0` could come from DSWx-HLS SAS and `1` from the PGE.
> 
> Any thoughts @hfattahi ?

@collinss-jpl and @gshiroma thanks for pointing this out. I agree with @gshiroma that the productVersion is not only the software version. I think "the algorithm software", "the inputs including ancillary" and "SDS version" should contribute to the product version. If the product is regenerated by a different SAS version, the different data version or different SDS version, I think the product version should be different. Of coarse, the algorithm/software will be Major and others will be minor or patch in a semantic versioning scheme.

hfattahi (approved):
Thank you @gshiroma and the team for your nice work.

gshiroma:
> 

Thank you, @hfattahi . This is very helpful. Let's bring back this discussion in our next meeting.

gshiroma:
@rtburns-jpl , we got 3 approvals, but the merging is still blocked. It seems that the issue is that reviewers don't have "write access." Any ideas?

rtburns-jpl:
Hmm, thank you for the heads up. I will look into this.

gshiroma:
> Hmm, thank you for the heads up. I will look into this.

Thank you, @rtburns-jpl

rtburns-jpl:
@hfattahi It looks like we're currently unable to merge this. I haven't heard back yet but I expect we will need to request write access, which might require a couple days to be approved. Is there an alternate way we can provide this delivery?

hfattahi:
@rtburns-jpl @gshiroma This is unfortunate. Is it possible to make a Docker image from Gustavo's branch?

hfattahi:
Could it be anything related to the PR wad originated from the repo in the opera-adt org before moving it to NASA org?
Can you guys open a dummy PR and see if we have permission for that to merge?

rtburns-jpl:
It's easy to make a docker image locally, just run the build_docker_image.sh script.

gshiroma:
@rtburns-jpl and @hfattahi , we created the Docker image locally. The biggest issue is the tagging that we provide in our documentation. We have two options, we either provide the tag as if existed or provide a tag from my own branch.

gshiroma:
Uhm, it looks like I don't have permissions to tag a release =/

gshiroma:
@rtburns-jpl , you probably would be able to do that (i.e. tag the release) for us. I guess you have write access, right? I wonder if there's anyone else on the team that also have write access so we can have two approvals and have this PR merged.

hfattahi:
For the documentation can we use PR#. let's clarify the issue we are facing in the release notes. Make it clear that this release is made from your branch because of lack of admin permission on this repo. As soon as we get permission on our repo we will make a quick point release

gshiroma:
> For the documentation can we use PR#. let's clarify the issue we are facing in the release notes. Make it clear that this release is made from your branch because of lack of admin permission on this repo. As soon as we get permission on our repo we will make a quick point release

Sounds good, @hfattahi . I just created this release: https://github.com/gshiroma/PROTEUS/releases/tag/v0.4 .  @jshimada47 , I think we'll have to use that release for the documents =/

jshimada47:
OK, I'll update the URLs in the documents.

Do you also want me to create an issue for product_version?  I'll just reference the comments in this PR.

gshiroma:
> OK, I'll update the URLs in the documents.
> 
> Do you also want me to create an issue for product_version? I'll just reference the comments in this PR.

Thank you, @jshimada47 . I wrote a draft note in the delivery summary. Please feel free to improve the text. Yes, that's a good idea to create the issue. I can do that now to save you time. Thanks!

bin/dswx_landcover_mask.py
gshiroma: Thank you, @nemo794 . I decided to remove this script `dswx_landcover_mask.py` from proteus. The script was basically a test script that I made public for John to test our landcover code.

src/proteus/dswx_hls.py
gshiroma: Very helpful, @nemo794 ! Thank you!

src/proteus/dswx_hls.py
```diff
@@ -2601,9 +2924,16 @@ def generate_dswx_layers(input_list,
         build_vrt_list.append(output_file)
         logger.info(f'file saved: {output_file}')
 
-    logger.info('list of output files:')
+    logger.info('removing temporary files:')
+    for filename in temp_files_list:
+        if not os.path.isfile(filename):
+            continue
+        # os.remove(filename)
+        logger.info(f'    {filename}')
```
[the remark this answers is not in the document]
gshiroma: Hi @nemo794 , I uncommented the `os.remove(filename)`. With respect to putting the code inside the `if` statement. I prefer the code as is. I prefer something like "early skipping" to reduce indentation levels. This piece of code here is small, so it doesn't make much difference, but I think if it were larger, the difference would be more clear.

src/proteus/dswx_hls.py
```diff
@@ -1874,7 +2076,8 @@ def get_projection_proj4(projection):
 def _relocate(input_file, geotransform, projection,
```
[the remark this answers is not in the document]
gshiroma: I'm renaming it to `_warp`. The function will reproject or relocate the raster depending on the parameter `projection`. Separating those two functionalities into two separate functions would make the code more complex and less efficient.

src/proteus/dswx_hls.py
```diff
@@ -2250,6 +2507,8 @@ def _compute_hillshade(dem_file, scratch_dir, sun_azimuth_angle,
     """
     shadow_layer_file = tempfile.NamedTemporaryFile(
         dir=scratch_dir, suffix='.tif').name
+    if temp_files_list is not None:
+        temp_files_list.append(shadow_layer_file)
```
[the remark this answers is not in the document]
gshiroma: `temp_files_list` is an optional list that can be provided to some functions so that the calling code knows which temporary files are being created.  The calling code is not required to have a list of temporary files. If `temp_files_list` is `None`, i.e. it's not provided, then nothing needs to be done.

src/proteus/dswx_hls.py
```diff
@@ -1888,39 +2091,47 @@ def _relocate(input_file, geotransform, projection,
        projection: str
               Output file's projection
        length: int
-              Output length
+              Output length before adding the margin defined by
+              `margin_in_pixels`
```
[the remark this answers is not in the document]
gshiroma: Thank you, @nemo794 . It's better to add the margins during relocation/reprojection. It would be very inefficient to add a separate step to do it.

src/proteus/dswx_hls.py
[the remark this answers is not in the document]
nemo794: Oh, I see. This function is akin to gdal's warp function. Sounds good.

src/proteus/dswx_hls.py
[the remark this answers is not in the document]
nemo794: Ah, thank you for clarifying.

However, my understanding is that temporary files should be removed before a program terminates. As-written, if a list is not provided, then there is no way for the garbage files to be cleaned up, leading to potentially huge storage space impacts. (For example the worldcover tmp file was often 100MB per granule. If I process 10 granules, that's an extra gig of storage space!)

As something to consider, if the tmp_files_list is not provided, it'd be good to have all temp files deleted automatically during this function. The doc string would then need to be updated to explain this behavior, e.g. "If tmp_files_list is provided, then the path to a tmp file will be appended to the list and not deleted. If tmp_files_list is not provided, then the tmp file will be deleted automatically."

At minimum, let's definitely update the docstring to reflect the current behavior. Maybe something like: 

```
...
       temp_files_list: list (optional)
              Mutable list of temporary files. If provided,
              paths to the temporary files generated will be
              appended to this list.

       Returns
       -------
       hillshade : numpy.ndarray
              Hillshade
              
       Notes
       ------
       All temporary files generated by this function will be
       placed in the `scratch_dir` and they will not be deleted.
       To delete, please supply an argument to `temp_files_list`.
       Paths to the temporary files generated will be appended
       to that list, which the user can use to find and delete the
       temp files.
```

src/proteus/dswx_hls.py
```diff
@@ -849,24 +904,22 @@ def _get_interpreted_dswx_ctable():
...
-    # Yellow - Reserved
-    dswx_ctable.SetColorEntry(7, (255, 255, 0))
-    # Orange - Reserved
-    dswx_ctable.SetColorEntry(8, (255, 127, 0))
+    if flag_collapse_wtr_classes:
+        # Green - Low confidence water or wetland
+        dswx_ctable.SetColorEntry(2, (0, 255, 0))
+    else:
+        # Light blue - Water (moderate conf.)
+        dswx_ctable.SetColorEntry(2, (0, 127, 255))
```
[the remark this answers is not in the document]
nemo794: Oh wow! This is a huge improvement! Makes it much easier to search for a variable and see how different values are being used throughout the code. Nice!

src/proteus/dswx_hls.py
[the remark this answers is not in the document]
nemo794: With the code as-written, I do not see how there would be a efficiency penalty for calling [these lines](https://github.com/gshiroma/PROTEUS/blob/fddc6aedbf18f8bf93c9f1dfdbce78e973f0a767/src/proteus/dswx_hls.py#L2276-L2290) ahead of time, and simply passing in the x/y coordinate start/stops.

Handling it separately actually mirrors gdal warp's API.

bin/dswx_landcover_mask.py
```diff
@@ -78,24 +88,126 @@ def _get_parser():
     return parser
 
 
+def point2epsg(lon, lat):
```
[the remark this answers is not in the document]
nemo794: For completeness, this script was removed entirely from the repo.

src/proteus/dswx_hls.py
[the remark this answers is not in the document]
gshiroma: Uhm. The file is created using `NamedTemporaryFile`. So, it should be deleted automatically after it's closed by PROTEUS:

"If delete is true (the default), the file is deleted as soon as it is closed. The returned object is always a file-like object whose file attribute is the underlying true file object. This file-like object can be used in a [with](https://docs.python.org/release/3.1.5/reference/compound_stmts.html#with) statement, just like a normal file." 

Source: https://docs.python.org/release/3.1.5/library/tempfile.html#tempfile.NamedTemporaryFile

Thank you for the suggestion of docstrings. I'll update them soon.

src/proteus/dswx_hls.py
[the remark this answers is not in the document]
gshiroma: I thought you wanted to add margins after the image is warped. You are suggesting to write a separate function just to add margin to the bounding box?

PR #33, opened 2022-12-02
Updated browse image and color for Partial Surface Water

nemo794:
This PR updates the browse image to represent the WTR-2 layer with "translucent" clouds and optional snow mask. By default, the snow mask will be applied; it is configurable via the runconfig and command line arguments to not apply the snow mask to the browse image.

For the collapsed water classification case, this PR also updates the partial surface water color in WTR, WTR-1, and WTR-2, and browse image to be light blue (RGB values of (180, 213, 244)).

An example output directory with all DSWx-HLS layers (except the 31 MB DEM.tif) is attached:
[test_output_dir.zip](https://github.com/nasa/PROTEUS/files/10143806/test_output_dir.zip)

To test, please run the standard dswx_hls.py workflow on any example dataset.


nemo794:
@hfattahi If you could please be a second reviewer on this PR, that would be much appreciated!

bin/dswx_hls.py
```diff
@@ -62,11 +62,12 @@ def main():
         output_shadow_masked_dswx=args.output_shadow_masked_dswx,
         output_landcover=args.output_landcover,
         output_shadow_layer=args.output_shadow_layer,
-        output_cloud_mask=args.output_cloud_mask,
+        output_cloud_mask=args.output_cloud_mask,
```
gshiroma, line 65:
```suggestion
        output_cloud_mask=args.output_cloud_mask,
```

src/proteus/dswx_hls.py
```diff
@@ -2355,6 +2480,105 @@ def _save_output_rgb_file(red, green, blue, output_file,
...
+        snow_idx = (cloud_layer == 2)
+        wtr2_clouds_layer[snow_idx] = WTR_CLOUD_MASKED_SNOW
+
+    # Mask the pixels with cloud and/or cloud shadow
+    cloud_idx = ((cloud_layer == 1) | 
+                (cloud_layer == 4) |
+                (cloud_layer == 5) |
+                (cloud_layer == 3) |
+                (cloud_layer == 6) |
+                (cloud_layer == 7))
```
gshiroma, line 2550: What if `flag_no_snow_mask` is `True`? I think, in that case, we should mimic the WTR class before we created the snow class 8, i.e., snow and clouds are represented by a single class "cloud+snow".

gshiroma: It's also better to put these conditions in order, i.e, `3` before `4`, and `5`.

src/proteus/dswx_hls.py
```diff
@@ -3437,47 +3670,61 @@ def generate_dswx_layers(input_list,
...
+                    flag_collapse_wtr_classes=FLAG_COLLAPSE_WTR_CLASSES,
+                    flag_no_snow_mask=flag_no_snow_mask_browse)
+
+        # Form color table
+        # Note: there is no need to account for `flag_no_snow_mask_browse` here.
+        # If `flag_no_snow_mask_browse` is False, then snow pixels would be set to
+        # `WTR_CLOUD_MASKED_SNOW` in _compute_wtr2_with_translucent_clouds_array(),
+        # and thus will appear as opaque snow.
+        # If `flag_no_snow_mask_browse` is True, then no pixels would be set to
+        # `WTR_CLOUD_MASKED_SNOW`, and thus no pixels would be colored as snow.
```
gshiroma, line 3696: Right, but I think they still should be masked as a single class cloud+snow (see comment above).

src/proteus/dswx_hls.py
```diff
@@ -3437,47 +3670,61 @@ def generate_dswx_layers(input_list,
...
+        wtr2_with_translucent_clouds_ctable = _get_wtr2_translucent_clouds_ctable(
+            alpha=0.35,
+            flag_collapse_wtr_classes=FLAG_COLLAPSE_WTR_CLASSES)
+
+        # Save to intermediary .tif temp file to scratch directory
+        tmp_filename = "dswx_wtr2_trans_clouds_tmp.tif"
```
gshiroma, line 3702: Please, use `tempfile.NamedTemporaryFile()` for that.

src/proteus/dswx_hls.py
```diff
@@ -2355,6 +2480,105 @@ def _save_output_rgb_file(red, green, blue, output_file,
...
+        wtr2_clouds_layer[ind] = OPEN_WATER_HIGH_CONF_AND_CLOUD
+
+        ind = ((wtr2_clouds_layer == WTR_UNCOLLAPSED_MODERATE_CONF_WATER) & cloud_idx)
+        wtr2_clouds_layer[ind] = OPEN_WATER_MOD_CONF_AND_CLOUD
+
+        ind = ((wtr2_clouds_layer == WTR_UNCOLLAPSED_POTENTIAL_WETLAND) & cloud_idx)
+        wtr2_clouds_layer[ind] = PARTIAL_WATER_HIGH_CONF_AND_CLOUD
+
+        ind = ((wtr2_clouds_layer == WTR_UNCOLLAPSED_LOW_CONF_WATER) & cloud_idx)
+        wtr2_clouds_layer[ind] = PARTIAL_WATER_AGGRESSIVE_AND_CLOUD
```
gshiroma, line 2577: These lines are way beyond 80 chars. Please, reorganize them.

src/proteus/dswx_hls.py
```diff
@@ -3349,7 +3581,7 @@ def generate_dswx_layers(input_list,
         invalid_ind = np.where(blue == image_dict['fill_data'])
     else:
         invalid_ind = np.where(np.isnan(blue))
-
+
```
gshiroma, line 3584:
```suggestion

```

src/proteus/dswx_hls.py
```diff
@@ -3437,47 +3670,61 @@ def generate_dswx_layers(input_list,
...
+        # Build the browse image
+        # Create the source image as a geotiff
+        # Reason: gdal.Create() cannot currently create .png files, so we
+        # must start from a GeoTiff, etc.
+        # Source: https://gis.stackexchange.com/questions/132298/gdal-c-api-how-to-create-png-or-jpeg-from-scratch
+        wtr2_with_translucent_clouds = _compute_wtr2_with_translucent_clouds_array(
+                    unmasked_interpreted_water_layer=landcover_shadow_masked_dswx,  # WTR-2
+                    cloud_layer=cloud,
+                    flag_collapse_wtr_classes=FLAG_COLLAPSE_WTR_CLASSES,
+                    flag_no_snow_mask=flag_no_snow_mask_browse)
```
gshiroma, line 3688: These lines are way too wide. This is not a strict rule, but please try to fit them within 80 chars.

gshiroma (commented):
Nice work, @nemo794 ! Thank you for working on that. I like how you implemented the logic for creating translucent clouds. I have a few minor comments below before approving it for merging.

src/proteus/dswx_hls.py
nemo794: > What if flag_no_snow_mask is True? I think, in that case, we should mimic the WTR class before we created the snow class 8, i.e., snow and clouds are represented by a single class "cloud+snow".

Ah, good point. Let's check with J. Jones and S. Chan?

As-is, the code is written this way to reflect my takeaway from the conversations: if the HLS FMask's encoding of snow is deemed to be faulty, then that would be the trigger to not apply the snow mask at all to those pixels, meaning that we should completely ignore the "snow" designation and not mask them.

That said, I could easily see that the takeaway you're suggesting here is actually what they meant. Totally happy to change; very easy fix.

Good question - thank you!

> It's also better to put these conditions in order, i.e, 3 before 4, and 5.
Ah - Good call.

src/proteus/dswx_hls.py
nemo794: Will do.

Out of curiosity, what is the reasoning behind having some temp files use `tempfile.NamedTemporaryFile()`, but these others having hardcoded names?

```
    ../scratch_dir/copernicus_reprojected.tif
    ../scratch_dir/worldcover_reprojected_up_3.tif
```

nemo794: Ahh, that makes sense. `tempfile` it is for all. Thank you for updating those other filenames!

hfattahi (commented):
Thank you Sam. This PR looks good to me. As we discussed offline, we need to keep the full res GTiff intermediate browse image as well. With that Change we are good to go.

gshiroma (approved):
LGTM. Thank you, @nemo794 !

nemo794:
Hi @gshiroma and @hfattahi , This PR is now updated per the emails this morning. The changes are fairly extensive, so it might be good to re-review.

Changes:

- Browse Image: snow and clouds both appear as translucent gray over WTR-2 layer
- Browse Image: PSW-Aggressive is by default not displayed. However, I added a flag in the runconfig so that it would be easy to display it again in the browse image in the future.
- Browse Image: Removed flag from runconfig to toggle snow on/off in browse image.
- Browse Image: Hardcoded RGB values for collapsed layers in browse; RGB values are per Project Scientist's PowerPoint slide.
- Updated `SOFTWARE_VERSION` to be 0.5.1. (@gshiroma , could you please check that I updated all necessary places?)
- Browse Image: high-res browse filename uses  `*_BROWSE.tif` and low-res browse filename uses  `*_BROWSE.png`

Here is a directory with the revised DSWx-HLS outputs. (Note: the DEM.tif was removed because the file is too big for git.)
[output_dir_v0.5.1.zip](https://github.com/nasa/PROTEUS/files/10233226/output_dir_v0.5.1.zip)

Thank you in advance for taking another look, and I look forward to your thoughts!

gshiroma:
Thank you for all updates, @nemo794 . This is really awesome. Since there are many changes, I'll remove the approval for now until I'm able to review this PR again.

gshiroma (commented):
@nemo794 , I liked your first implementation of this code, but I think I don't like the recent changes. Particularly, the way you are treating the snow class is making the code unnecessarily complex, in my opinion. It's now difficult to follow the number of constants and if-else statements. As we discussed, a simpler option is to apply the three snow options at the moment that we convert WTR-2 into WTR. That would be a single point of change. If you are too worried about adding a snow class 8 to the color table when "it's not needed," you can skip that class when creating the color table. This solution would also apply to WTR, which I believe is what is being evaluated by the project, as the merging of snow and cloud classes would happen at the WTR level as well. Please, let me know if I'm missing something.

nemo794:
> @nemo794 , I liked your first implementation of this code, but I think I don't like the recent changes. Particularly, the way you are treating the snow class is making the code unnecessarily complex, in my opinion. It's now difficult to follow the number of constants and if-else statements. As we discussed, a simpler option is to apply the three snow options at the moment that we convert WTR-2 into WTR. That would be a single point of change. If you are too worried about adding a snow class 8 to the color table when "it's not needed," you can skip that class when creating the color table. This solution would also apply to WTR, which I believe is what is being evaluated by the project, as the merging of snow and cloud classes would happen at the WTR level as well. Please, let me know if I'm missing something.

Hi. @gshiroma , while I agree in principle that the WTR and Browse Image should probably be handled similarly, at this time, the directive from the project is only to update the Browse Image for this PR. The project did not say anything about updating the WTR layer (except for the two-blues color scheme), and thus this PR does not touch that code.

Also, please remember that WTR and BROWSE are created independently; WTR is not an input to BROWSE, aka BROWSE is not dependent upon WTR. If there is directive from project to update WTR at a later time, then the runconfig flag can be renamed and moved into the runconfig's `processing_group`, and that single parameter can then be piped to each of WTR and BROWSE.

hfattahi (approved):
LGTM. Thanks @nemo794

gshiroma:
> > @nemo794 , I liked your first implementation of this code, but I think I don't like the recent changes. Particularly, the way you are treating the snow class is making the code unnecessarily complex, in my opinion. It's now difficult to follow the number of constants and if-else statements. As we discussed, a simpler option is to apply the three snow options at the moment that we convert WTR-2 into WTR. That would be a single point of change. If you are too worried about adding a snow class 8 to the color table when "it's not needed," you can skip that class when creating the color table. This solution would also apply to WTR, which I believe is what is being evaluated by the project, as the merging of snow and cloud classes would happen at the WTR level as well. Please, let me know if I'm missing something.
> 
> Hi. @gshiroma , while I agree in principle that the WTR and Browse Image should probably be handled similarly, at this time, the directive from the project is only to update the Browse Image for this PR. The project did not say anything about updating the WTR layer (except for the two-blues color scheme), and thus this PR does not touch that code.
> 
> Also, please remember that WTR and BROWSE are created independently; WTR is not an input to BROWSE, aka BROWSE is not dependent upon WTR. If there is directive from project to update WTR at a later time, then this new runconfig flag can be renamed and moved into the runconfig's `processing_group`, and that single parameter can then be piped to each of WTR and BROWSE.

As far as I understood, the request also involved WTR. You probably remembered the discussion we had with John. In my opinion, this should be clarified before moving forward to avoid losing unnecessary time. 

In any case, if you really don't want to change WTR (which I think we should), you can still update the browse code using the same idea. For example, you could change the function `_compute_wtr2_translucent_clouds_array()` in the line that you make the assignment `snow_idx = (cloud_layer == 2)`. If you consider snow and cloud the same class, you can make`snow_idx` "empty". You can keep your code `if set_snow_to_nodata` then `wtr2_clouds[snow_idx] = UINT8_FILL_VALUE`. These two pieces of code should handle all three options without any changes to `_get_wtr2_translucent_clouds_ctable()`. Please, let me know if the idea is not clear, we can also discuss that in a short meeting if it's helpful.

src/proteus/dswx_hls.py
```diff
@@ -2539,44 +2930,44 @@ def parse_runconfig_file(user_runconfig_file = None, args = None):
     else:
         runconfig = default_runconfig
 
-    runconfig_constants = RunConfigConstants()
+    runcfg_consts = RunConfigConstants()
```
gshiroma, line 2933: BTW, I'd prefer to keep `runconfig_constants` rather than `runcfg_consts`. I think the first former is easier to read, and the difference in variable name length is not that big to justify the change. Would you be OK with keeping it as `runconfig_constants`?

nemo794: Either way is fine with me. The variable name was shortened in response to the PR comment (https://github.com/nasa/PROTEUS/pull/33#discussion_r1039012166) asking lines remain under 80 characters. By using shorter variable names, it made it easier to keep several lines under that limit.

Would you prefer to have a longer variable name, and more lines split onto two lines?

nemo794:
Updated Browse Image per

> > > @nemo794 , I liked your first implementation of this code, but I think I don't like the recent changes. Particularly, the way you are treating the snow class is making the code unnecessarily complex, in my opinion. It's now difficult to follow the number of constants and if-else statements. As we discussed, a simpler option is to apply the three snow options at the moment that we convert WTR-2 into WTR. That would be a single point of change. If you are too worried about adding a snow class 8 to the color table when "it's not needed," you can skip that class when creating the color table. This solution would also apply to WTR, which I believe is what is being evaluated by the project, as the merging of snow and cloud classes would happen at the WTR level as well. Please, let me know if I'm missing something.
> > 
> > 
> > Hi. @gshiroma , while I agree in principle that the WTR and Browse Image should probably be handled similarly, at this time, the directive from the project is only to update the Browse Image for this PR. The project did not say anything about updating the WTR layer (except for the two-blues color scheme), and thus this PR does not touch that code.
> > Also, please remember that WTR and BROWSE are created independently; WTR is not an input to BROWSE, aka BROWSE is not dependent upon WTR. If there is directive from project to update WTR at a later time, then this new runconfig flag can be renamed and moved into the runconfig's `processing_group`, and that single parameter can then be piped to each of WTR and BROWSE.
> 
> As far as I understood, the request also involved WTR. You probably remembered the discussion we had with John. In my opinion, this should be clarified before moving forward to avoid losing unnecessary time.
> 
> In any case, if you really don't want to change WTR (which I think we should), you can still update the browse code using the same idea. For example, you could change the function `_compute_wtr2_translucent_clouds_array()` in the line that you make the assignment `snow_idx = (cloud_layer == 2)`. If you consider snow and cloud the same class, you can make`snow_idx` "empty". You can keep your code `if set_snow_to_nodata` then `wtr2_clouds[snow_idx] = UINT8_FILL_VALUE`. These two pieces of code should handle all three options without any changes to `_get_wtr2_translucent_clouds_ctable()`. Please, let me know if the idea is not clear, we can also discuss that in a short meeting if it's helpful.

It's not that I don't want to change WTR; it's that that was outside of the scope of this PR, and the current direction we're receiving from the project is only concerning BROWSE. When new direction is provided for WTR, then let's definitely circle back to this.

nemo794:
Hi @gshiroma and @hfattahi Updated PR per the discussion today to

nemo794:
Hi @gshiroma and @hfattahi ,

The Browse Image PR is updated per today's discussion. All of the "translucent" code was stripped out, making the code much simpler overall. Also, the Browse Image now relies on only the WTR layer, and no longer on CLOUD nor WTR-2.

Here's what was added:

* `snow_in_browse` flag in runconfig now has these three options: opaque cyan, opaque gray, and nodata
* `cloud_in_browse` flag in runconfig has two options: opaque gray and nodata

Here are example outputs (without DEM), showing the six configurations:
[tahoe_CloudGray_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10274229/tahoe_CloudGray_SnowCyan.zip)
[tahoe_CloudGray_SnowGray.zip](https://github.com/nasa/PROTEUS/files/10274230/tahoe_CloudGray_SnowGray.zip)
[tahoe_CloudGray_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10274231/tahoe_CloudGray_SnowNodata.zip)
[tahoe_CloudNodata_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10274232/tahoe_CloudNodata_SnowCyan.zip)
[tahoe_CloudNodata_SnowGray.zip](https://github.com/nasa/PROTEUS/files/10274234/tahoe_CloudNodata_SnowGray.zip)
[tahoe_CloudNodata_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10274235/tahoe_CloudNodata_SnowNodata.zip)

@gshiroma , the only item remaining is what to name the runconfig constants variable; my reason for shortening the variable name is noted [here](https://github.com/nasa/PROTEUS/pull/33#discussion_r1053934854), but either way is dandy. Please let me know what you'd prefer!

Thank you!

hfattahi:
> Hi @gshiroma and @hfattahi ,
> 
> The Browse Image PR is updated per today's discussion. All of the "translucent" code was stripped out, making the code much simpler overall. Also, the Browse Image now relies on only the WTR layer, and no longer on CLOUD nor WTR-2.
> 
> Here's what was added:
> 
> * `snow_in_browse` flag in runconfig now has these three options: opaque cyan, opaque gray, and nodata
> * `cloud_in_browse` flag in runconfig has two options: opaque gray and nodata
> 
> Here are example outputs (without DEM), showing the six configurations: [tahoe_CloudGray_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10274229/tahoe_CloudGray_SnowCyan.zip) [tahoe_CloudGray_SnowGray.zip](https://github.com/nasa/PROTEUS/files/10274230/tahoe_CloudGray_SnowGray.zip) [tahoe_CloudGray_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10274231/tahoe_CloudGray_SnowNodata.zip) [tahoe_CloudNodata_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10274232/tahoe_CloudNodata_SnowCyan.zip) [tahoe_CloudNodata_SnowGray.zip](https://github.com/nasa/PROTEUS/files/10274234/tahoe_CloudNodata_SnowGray.zip) [tahoe_CloudNodata_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10274235/tahoe_CloudNodata_SnowNodata.zip)
> 
> @gshiroma , the only item remaining is what to name the runconfig constants variable; my reason for shortening the variable name is noted [here](https://github.com/nasa/PROTEUS/pull/33#discussion_r1053934854), but either way is dandy. Please let me know what you'd prefer!
> 
> Thank you!

Thanks @nemo794 this is great and covers most of what we discussed. The last remaining piece is no-water class to have two options:
- white
- no-data

Would you please add that as well?

nemo794:
> Thanks @nemo794 this is great and covers most of what we discussed. The last remaining piece is no-water class to have two options:
> 
> * white
> * no-data
> 
> Would you please add that as well?

Thanks, @hfattahi for catching that. This additional option for Not Water has been added. Here are a few examples of the updated output datasets (sans *DEM.tif):
[tahoe_NotwaterNoData_CloudNodata_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10280940/tahoe_NotwaterNoData_CloudNodata_SnowNodata.zip)
[tahoe_NotwaterWhite_CloudGray_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10280943/tahoe_NotwaterWhite_CloudGray_SnowNodata.zip)
[tahoe_NotwaterWhite_CloudGray_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10280945/tahoe_NotwaterWhite_CloudGray_SnowCyan.zip)
[tahoe_NotwaterNoData_CloudGray_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10280947/tahoe_NotwaterNoData_CloudGray_SnowCyan.zip)

src/proteus/dswx_hls.py
nemo794: Per our offline discussion, the variable name was changed back to `runconfig_constants ` from from `runcfg_consts`.

hfattahi:
> > Thanks @nemo794 this is great and covers most of what we discussed. The last remaining piece is no-water class to have two options:
> > 
> > * white
> > * no-data
> > 
> > Would you please add that as well?
> 
> Thanks, @hfattahi for catching that. This additional option for Not Water has been added. Here are a few examples of the updated output datasets (sans *DEM.tif): [tahoe_NotwaterNoData_CloudNodata_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10280940/tahoe_NotwaterNoData_CloudNodata_SnowNodata.zip) [tahoe_NotwaterWhite_CloudGray_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10280943/tahoe_NotwaterWhite_CloudGray_SnowNodata.zip) [tahoe_NotwaterWhite_CloudGray_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10280945/tahoe_NotwaterWhite_CloudGray_SnowCyan.zip) [tahoe_NotwaterNoData_CloudGray_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10280947/tahoe_NotwaterNoData_CloudGray_SnowCyan.zip)

Thanks @nemo794 . Would you please double check the *BROWSE.tif (high res browse) for [tahoe_NotwaterNoData_CloudNodata_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10280940/tahoe_NotwaterNoData_CloudNodata_SnowNodata ? Why the lake has disappeared?

nemo794:
> > > Thanks @nemo794 this is great and covers most of what we discussed. The last remaining piece is no-water class to have two options:
> > > 
> > > * white
> > > * no-data
> > > 
> > > Would you please add that as well?
> > 
> > 
> > Thanks, @hfattahi for catching that. This additional option for Not Water has been added. Here are a few examples of the updated output datasets (sans *DEM.tif): [tahoe_NotwaterNoData_CloudNodata_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10280940/tahoe_NotwaterNoData_CloudNodata_SnowNodata.zip) [tahoe_NotwaterWhite_CloudGray_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10280943/tahoe_NotwaterWhite_CloudGray_SnowNodata.zip) [tahoe_NotwaterWhite_CloudGray_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10280945/tahoe_NotwaterWhite_CloudGray_SnowCyan.zip) [tahoe_NotwaterNoData_CloudGray_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10280947/tahoe_NotwaterNoData_CloudGray_SnowCyan.zip)
> 
> Thanks @nemo794 . Would you please double check the *BROWSE.tif (high res browse) for [tahoe_NotwaterNoData_CloudNodata_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10280940/tahoe_NotwaterNoData_CloudNodata_SnowNodata ? Why the lake has disappeared?

I noticed a similar thing happening with CLOUD layer a couple of weeks ago when I tried opening that tif in Preview, and showed it to @gshiroma . In that scenario, the colortable was correct, and the .tif opened/displayed correctly in Google Earth Pro with the correct color scheme, etc., so we figured it was a quirk of Preview. I dug into the code history a bit, and could not find any changes that seemed relevant. Strangely, still without changing the code, the CLOUD issue has resolved itself.

Looking at this particular *BROWSE.tif file, it is exhibiting the same behavior. The colortable looks correct, dimensions are correct, and it opens in Google Earth Pro with the water visualized (screenshot below). It is also the exact file used to create the *BROWSE.png, which looks correct. But, when opening in Preview, it appears completely transparent (not correct).

I skimmed through a random assortment of old outputs, and this issue seems to appear intermittently in other layers (WTR, WTR-2) as well. No changes to those portions of the code. What's curious is that the incorrect layers have a size in the 100-200KB range, whereas the correct layers are in the >500 KB range.

If this is something we should look into farther, please say the word. But due to its intermittent behavior and impact to various layers, it might be best to raise it as a new Issue.

Screen Shot 2022-12-21 at 3 08 35 PM

hfattahi:
> > > > Thanks @nemo794 this is great and covers most of what we discussed. The last remaining piece is no-water class to have two options:
> > > > 
> > > > * white
> > > > * no-data
> > > > 
> > > > Would you please add that as well?
> > > 
> > > 
> > > Thanks, @hfattahi for catching that. This additional option for Not Water has been added. Here are a few examples of the updated output datasets (sans *DEM.tif): [tahoe_NotwaterNoData_CloudNodata_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10280940/tahoe_NotwaterNoData_CloudNodata_SnowNodata.zip) [tahoe_NotwaterWhite_CloudGray_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10280943/tahoe_NotwaterWhite_CloudGray_SnowNodata.zip) [tahoe_NotwaterWhite_CloudGray_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10280945/tahoe_NotwaterWhite_CloudGray_SnowCyan.zip) [tahoe_NotwaterNoData_CloudGray_SnowCyan.zip](https://github.com/nasa/PROTEUS/files/10280947/tahoe_NotwaterNoData_CloudGray_SnowCyan.zip)
> > 
> > 
> > Thanks @nemo794 . Would you please double check the *BROWSE.tif (high res browse) for [tahoe_NotwaterNoData_CloudNodata_SnowNodata.zip](https://github.com/nasa/PROTEUS/files/10280940/tahoe_NotwaterNoData_CloudNodata_SnowNodata ? Why the lake has disappeared?
> 
> I noticed a similar thing happening with CLOUD layer a couple of weeks ago when I tried opening that tif in Preview, and showed it to @gshiroma . In that scenario, the colortable was correct, and the .tif opened/displayed correctly in Google Earth Pro with the correct color scheme, etc., so we figured it was a quirk of Preview. I dug into the code history a bit, and could not find any changes that seemed relevant. Strangely, still without changing the code, the CLOUD issue has resolved itself.
> 
> Looking at this particular *BROWSE.tif file, it is exhibiting the same behavior. The colortable looks correct, dimensions are correct, and it opens in Google Earth Pro with the water visualized (screenshot below). It is also the exact file used to create the *BROWSE.png, which looks correct. But, when opening in Preview, it appears completely transparent (not correct).
> 
> I skimmed through a random assortment of old outputs, and this issue seems to appear intermittently in other layers (WTR, WTR-2) as well. No changes to those portions of the code. What's curious is that the incorrect layers have a size in the 100-200KB range, whereas the correct layers are in the >500 KB range.
> 
> If this is something we should look into farther, please say the word. But due to its intermittent behavior and impact to various layers, it might be best to raise it as a new Issue.
> 
> Screen Shot 2022-12-21 at 3 08 35 PM

Thanks @nemo794 . You are right I can visualize it different ways. So looks good to me.

gshiroma (commented):
Great job, @nemo794 ! The code is clear, clean, and well-documented. That's sad that you had to remove the "translucent" option, I kind of liked that. Anyways, I have a few comments below that we should review before merging this PR. Thanks again!

src/proteus/dswx_hls.py
```diff
@@ -1118,11 +1178,94 @@ def _get_interpreted_dswx_ctable(
...
+        out_ctable.SetColorEntry(WTR_CLOUD_MASKED_SNOW, cloud_gray)
+    elif snow_color == 'nodata':
+        # The no-data fill RGBA was set by `_get_interpreted_dswx_ctable`.
+        # So, "remove" original snow entry from colortable.
+        out_ctable.SetColorEntry(WTR_CLOUD_MASKED_SNOW, FILL_VALUE_RGBA)
+    else:
+        # Snow color was already set to cyan in `_get_interpreted_dswx_ctable`.
+        # Perform sanity check in case of changes.
+        assert out_ctable.GetColorEntry(WTR_CLOUD_MASKED_SNOW) in \
+                ((0,255,255), (0,255,255,255)), "Color for snow has changed."
```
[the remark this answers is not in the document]
nemo794: Good point. I'll go ahead and remove these, but leave comments re: why we're not handling all cases.

src/proteus/dswx_hls.py
```diff
@@ -2355,6 +2498,74 @@ def _save_output_rgb_file(red, green, blue, output_file,
...
+    # Create a copy of the masked_interpreted_water_layer.
+    browse_arr = masked_interpreted_water_layer.copy()
+
+    # Discard the Partial Surface Water Aggressive class
+    if not include_psw_aggressive and flag_collapse_wtr_classes:
+        browse_arr[browse_arr == WTR_UNCOLLAPSED_LOW_CONF_WATER] = \
+                                                            WTR_NOT_WATER
+
+    if flag_collapse_wtr_classes:
+        browse_arr = _collapse_wtr_classes(browse_arr)
```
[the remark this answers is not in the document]
nemo794: The code actually uses `_save_array()` to save the browse image; it does not use `save_dswx_product()`. The browse image uses `_save_arr()` because it needs a custom color table.

nemo794:
Thank you @hfattahi and @gshiroma for reviewing! All really insightful and helpful comments.

I've addressed the notes. Also, per the offline conversation, I've changed the default behavior so that PSW-Aggressive class is displayed in the Browse Image. To make this clearer for the user (particularly is using CLI), the parameter name in the runconfig was updated to `exclude_psw_aggressive_in_browse`.

src/proteus/dswx_hls.py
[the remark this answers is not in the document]
gshiroma: Thank you, @nemo794 . I'm not a fan of `else: pass`, but I guess it's ok (for clarity).

src/proteus/dswx_hls.py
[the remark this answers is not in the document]
gshiroma: Oh, I see. Good point. I forgot to mention that you may want to add a description to the browse .tif file. But we can do that later too (in another PR).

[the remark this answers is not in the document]
nemo794: Sounds good.

And if we remember to add the description, that's fine. But the BROWSE.tif is only being used by DAAC for creating add'l images, and it will not be included in the final granules for the users, so it's probably not necessary?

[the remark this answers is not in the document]
gshiroma: Yes, it's not necessary, but a small thing that would be nice to add at some point, doesn't need to be now.

nemo794:
Confirmed with @gshiroma , I will merge this PR even though the CircleCI build is failing. It is failing due to incorrect golden datasets. After merging this PR, the changes will be incorporated into PR #32 which is nearly ready to merge. The new golden dataset will then be created based on that PR, after which the CircleCI build should pass.

PR #37, opened 2023-01-27
upload the defaults and schema for dswx_s1

oberonia78:
This PR adds the schema and defaults for DSWx-S1 algorithm. Since this is the first PR, we may need to discuss what way is the most appropriate for DSWx-HLS and DSWx-S1.


oberonia78:
No more needed. The DSWx-SAR will be in the independent repository.

PR #38, opened 2023-01-29
Update proteus for release v1.0

gshiroma:
This PR updates proteus for release v1.0. Changes:
- Rename metadata fields imported from HLS product from `SPATIAL_COVERAGE` and `CLOUD_COVERAGE` to `INPUT_HLS_PRODUCT_SPATIAL_COVERAGE` and `INPUT_HLS_PRODUCT_CLOUD_COVERAGE`, respectively;
- Recompute and store the new metadata fields `SPATIAL_COVERAGE` and `CLOUD_COVERAGE` considering fill value and ocean masked data as invalid data;
- Save new metadata fields describing processing parameters used for generating the DSWx-HLS product: `SHADOW_MASKING_ALGORITHM`, `MIN_SLOPE_ANGLE`, `MAX_SUN_LOCAL_INC_ANGLE`, `MASK_ADJACENT_TO_CLOUD_MODE`, `FOREST_MASK_LANDCOVER_CLASSES`, and `OCEAN_MASKING_SHORELINE_DISTANCE_KM`;
- Update version tag from `v.0.5.2` to `v1.0`.
- Add high-aerosol masking.

### DIFF
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -1149,7 +1149,7 @@ def _is_landcover_class_high_intensity_developed(landcover_mask):
     return high_intensity_developed_mask
 
 
-def _apply_aerosol_class_remapping_wtr1_class(wtr_1_layer, nir,
+def _apply_aerosol_class_remapping_single_class(wtr_1_layer, nir,
         preliminary_cloud_layer, fmask,
         fmask_values, input_wtr1_class, output_wtr1_class):
     """Apply aerosol remapping onto interpreted layer (WTR-1) given a
@@ -1176,14 +1176,9 @@ def _apply_aerosol_class_remapping_wtr1_class(wtr_1_layer, nir,
             Remapped WTR-1 class
     """
 
-    to_remap_array = None
-    for fmask_value in fmask_values:
-        if to_remap_array is None:
-            to_remap_array = fmask == fmask_value
-        else:
-            to_remap_array |= fmask == fmask_value
-    to_remap_array &= wtr_1_layer == input_wtr1_class
-    to_remap_array &= nir <= AEROSOL_REMAPPING_MAX_NIR
+    to_remap_array = ((np.isin(fmask, fmask_values)) &
+                      (wtr_1_layer == input_wtr1_class) &
+                      (nir <= AEROSOL_REMAPPING_MAX_NIR))
     wtr_1_layer[to_remap_array] = output_wtr1_class
 
     # set CLOUD layer bit (3): 2**3 = 8
@@ -1244,7 +1239,7 @@ def _apply_aerosol_class_remapping(wtr_1_layer, nir,
 
     for input_wtr1_class, (fmask_values, output_wtr1_class) in \
             wtr1_class_fmask_values_dict.items():
-        _apply_aerosol_class_remapping_wtr1_class(wtr_1_layer,
+        _apply_aerosol_class_remapping_single_class(wtr_1_layer,
             nir, preliminary_cloud_layer, fmask,
             fmask_values, input_wtr1_class, output_wtr1_class)
 
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -4315,7 +4315,7 @@ def generate_dswx_layers(input_list,
 
     # use floor integer to compute spatial and cloud coverage
     spatial_coverage = int(100 * float(n_valid) / total_number_of_pixels)
-    cloud_coverage = int(100 * float(n_cloud_and_valid) / total_number_of_pixels)
+    cloud_coverage = int(100 * float(n_cloud_and_valid) / n_valid)
 
     # print spatial and cloud coverage
     logger.info(f'    spatial coverage [%]:  {spatial_coverage}')
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -1713,14 +1713,14 @@ def _compute_preliminary_cloud_layer(fmask, mask_adjacent_to_cloud_mode):
 
        Returns
        -------
-            preliminary_cloud_layer : numpy.ndarray
-                Preliminary cloud mask (without the snow/ice class)
-                with dtype of uint8 and values assigned as follows:
-                0: Not masked
-                1: Cloud shadow or adjacent to cloud/cloud shadow
-                4: Cloud
-                5: Cloud and class 1 (cloud shadow or adjacent to cloud/cloud shadow)
-                255: Fill value (no data)
+       preliminary_cloud_layer : numpy.ndarray
+           Preliminary cloud mask (without the snow/ice class)
+           with dtype of uint8 and values assigned as follows:
+           0: Not masked
+           1: Cloud shadow or adjacent to cloud/cloud shadow
+           4: Cloud
+           5: Cloud and class 1 (cloud shadow or adjacent to cloud/cloud shadow)
+           255: Fill value (no data)
 
        See Also
        ---------
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -1496,7 +1496,9 @@ def _get_cloud_layer_ctable():
     mask_ctable.SetColorEntry(7, (127, 127, 255))
 
     # Light ocre - Aerosol reassignment ("0xE4CDA7")
-    mask_ctable.SetColorEntry(8, (228, 205, 167))
+    # mask_ctable.SetColorEntry(8, (228, 205, 167))
+    # Grayish blue - Aerosol reassignment ("0x7393B3")
+    mask_ctable.SetColorEntry(8, (115, 147, 179))
     # Dark gray - Cloud shadow
     mask_ctable.SetColorEntry(9, (64, 64, 64))
     # Cyan - snow/ice
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -1154,9 +1154,13 @@ def _apply_aerosol_masking_wtr1_class(wtr_1_layer, fmask, wtr1_class,
             in the presence of high aerosol
     """
 
-    to_mask_array = wtr_1_layer == wtr1_class
+    to_mask_array = None
     for fmask_value in fmask_values:
-        to_mask_array &= fmask == fmask_value
+        if to_mask_array is None:
+            to_mask_array = fmask == fmask_value
+        else:
+            to_mask_array |= fmask == fmask_value
+    to_mask_array &= wtr_1_layer == wtr1_class
     wtr_1_layer[to_mask_array] = WATER_UNCOLLAPSED_HIGH_CONF_CLEAR
 
 
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -1195,7 +1195,7 @@ def _apply_aerosol_masking(wtr_1_layer, fmask,
     wtr1_class_fmask_values_dict = {
         WATER_NOT_WATER_CLEAR:
             (aerosol_not_water_to_high_conf_water_fmask_values,
-             WATER_UNCOLLAPSED_HIGH_CONF_CLEAR),
+             WATER_UNCOLLAPSED_MODERATE_CONF_CLEAR),
         WATER_UNCOLLAPSED_MODERATE_CONF_CLEAR:
             (aerosol_water_moderate_conf_to_high_conf_water_fmask_values,
              WATER_UNCOLLAPSED_HIGH_CONF_CLEAR),
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -3719,7 +3719,8 @@ def _compute_hillshade(dem_file, scratch_dir, sun_azimuth_angle,
 def _compute_opera_shadow_layer(dem, sun_azimuth_angle, sun_elevation_angle,
                                 min_slope_angle, max_sun_local_inc_angle,
                                 pixel_spacing_x = 30, pixel_spacing_y = 30):
-    """Compute hillshade using new OPERA shadow masking
+    """Compute shadow mask based on Sun local incidence angle and slope
+       angle.
 
        Parameters
        ----------
@@ -3742,8 +3743,8 @@ def _compute_opera_shadow_layer(dem, sun_azimuth_angle, sun_elevation_angle,
 
        Returns
        -------
-       hillshade : numpy.ndarray
-              Hillshade
+       shadow_mask : numpy.ndarray
+              Shadow mask (1: not shadow, 0: shadow)
     """
     sun_azimuth = np.radians(sun_azimuth_angle)
     sun_zenith_degrees = 90 - sun_elevation_angle
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -315,20 +315,20 @@ class RunConfigConstants:
         HLS reflectance thresholds for generating DSWx-HLS products
     check_ancillary_inputs_coverage: bool
         Check if ancillary inputs cover entirely the output product
-    apply_aerosol_masking: bool
+    apply_aerosol_class_remapping: bool
         Apply aerosol masking
-    aerosol_not_water_to_high_conf_water_fmask_values: list(int)
-         HLS Fmask values to convert not-water to high-confidence water
+    aerosol_not_water_to_moderate_conf_water_fmask_values: list(int)
+         HLS Fmask values to convert not-water to moderate-confidence water
          in the presence of high aerosol
     aerosol_water_moderate_conf_to_high_conf_water_fmask_values: list(int)
          HLS Fmask values to convert moderate-confidence water to
          high-confidence water in the presence of high aerosol
-    aerosol_partial_surface_water_conservative_to_high_conf_water_fmask_values: list(int)
+    aerosol_partial_surface_water_conservative_to_moderate_conf_water_fmask_values: list(int)
          HLS Fmask values to convert partial surface water conservative to
-         high-confidence water in the presence of high aerosol
-    aerosol_partial_surface_aggressive_to_high_conf_water_fmask_values: list(int)
+         moderate-confidence water in the presence of high aerosol
+    aerosol_partial_surface_aggressive_to_moderate_conf_water_fmask_values: list(int)
          HLS Fmask values to convert partial surface water aggressive to
-         high-confidence water in the presence of high aerosol
+         moderate-confidence water in the presence of high aerosol
     shadow_masking_algorithm: str
         Shadow masking algorithm
     min_slope_angle: float
@@ -373,11 +373,11 @@ class RunConfigConstants:
     def __init__(self):
         self.hls_thresholds = HlsThresholds()
         self.check_ancillary_inputs_coverage = None
-        self.apply_aerosol_masking = None
-        self.aerosol_not_water_to_high_conf_water_fmask_values = None
+        self.apply_aerosol_class_remapping = None
+        self.aerosol_not_water_to_moderate_conf_water_fmask_values = None
         self.aerosol_water_moderate_conf_to_high_conf_water_fmask_values = None
-        self.aerosol_partial_surface_water_conservative_to_high_conf_water_fmask_values = None
-        self.aerosol_partial_surface_aggressive_to_high_conf_water_fmask_values = None
+        self.aerosol_partial_surface_water_conservative_to_moderate_conf_water_fmask_values = None
+        self.aerosol_partial_surface_aggressive_to_moderate_conf_water_fmask_values = None
         self.shadow_masking_algorithm = None
         self.min_slope_angle = None
         self.max_sun_local_inc_angle = None
@@ -618,7 +618,7 @@ def get_dswx_hls_cli_parser():
                               ' the output product'))
 
     parser.add_argument('--apply-aerosol-masking',
-                        dest='apply_aerosol_masking',
+                        dest='apply_aerosol_class_remapping',
                         action='store_true',
                         default=None,
                         help='Apply aerosol masking')
@@ -1135,7 +1135,7 @@ def _is_landcover_class_high_intensity_developed(landcover_mask):
     return high_intensity_developed_mask
 
 
-def _apply_aerosol_masking_wtr1_class(wtr_1_layer, fmask,
+def _apply_aerosol_class_remapping_wtr1_class(wtr_1_layer, fmask,
         fmask_values, input_wtr1_class, output_wtr1_class):
     """Apply aerosol masking onto interpreted layer (WTR-1) given a
        WTR-1 class and fmask values
@@ -1147,12 +1147,12 @@ def _apply_aerosol_masking_wtr1_class(wtr_1_layer, fmask,
        fmask: numpy.ndarray
             HLS Fmask
        fmask_values: list(int)
-            HLS Fmask values to convert not-water to high-confidence water
+            HLS Fmask values to remap interpreted water classes
             in the presence of high aerosol
        input_wtr1_class: int
             Input WTR-1 class that is being evaluated for high-aerosol
        output_wtr1_class: int
-            Output WTR-1 class if input WTR-1 class has high-aerosol
+            Remapped WTR-1 class
     """
 
     to_mask_array = None
@@ -1165,11 +1165,11 @@ def _apply_aerosol_masking_wtr1_class(wtr_1_layer, fmask,
     wtr_1_layer[to_mask_array] = output_wtr1_class
 
 
-def _apply_aerosol_masking(wtr_1_layer, fmask,
-        aerosol_not_water_to_high_conf_water_fmask_values,
+def _apply_aerosol_class_remapping(wtr_1_layer, fmask,
+        aerosol_not_water_to_moderate_conf_water_fmask_values,
         aerosol_water_moderate_conf_to_high_conf_water_fmask_values,
-        aerosol_partial_surface_water_conservative_to_high_conf_water_fmask_values,
-        aerosol_partial_surface_aggressive_to_high_conf_water_fmask_values):
+        aerosol_partial_surface_water_conservative_to_moderate_conf_water_fmask_values,
+        aerosol_partial_surface_aggressive_to_moderate_conf_water_fmask_values):
     """Apply aerosol masking onto interpreted layer (WTR-1)
 
        Parameters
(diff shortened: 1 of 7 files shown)


src/proteus/dswx_hls.py
```diff
@@ -4222,11 +4255,52 @@ class in the browse image. False to display PSW-Agg.
...
+
+    if 'INPUT_HLS_PRODUCT_CLOUD_COVERAGE' in dswx_metadata_dict.keys():
+        hls_product_cloud_coverage = dswx_metadata_dict[
+            'INPUT_HLS_PRODUCT_CLOUD_COVERAGE']
+        logger.info(f'    input HLS product cloud coverage [%]:'
+                    f' {hls_product_cloud_coverage}')
+
+    # compute a preliminary cloud layer, i.e., without snow/ice class
+    preliminary_cloud_layer = _compute_preliminary_cloud_layer(
+        fmask, mask_adjacent_to_cloud_mode)
```
nemo794, line 4274: I've gone back and forth about this preliminary cloud layer. It was really confusing to me until I spent a bit of time with the code, and I really do not like splitting the logic for the CLOUD layer into two separate places; the docstrings are confusing/redundant, and creates extra steps that need to be tracked and followed up on later in the code.

What about a new approach: 

1) Put all logic for creating the CLOUD layer (cloud+shadow+adj+snow) into a new function, e.g. `_compute_cloud_layer()`.
2) Call that function at this point in the code.
3) When computing the statistic of `CLOUD_COVERAGE`, it would become something like:
```
n_cloud_and_valid = np.sum(((preliminary_cloud_layer) != 0) & 
                                                    (preliminary_cloud_layer) != SNOW) & 
                                                    (valid_array))
```
4) Later, turn the function `_compute_and_apply_cloud_layer()` into simply `_compute_wtr_2_layer(wtr_1_layer=..., cloud_layer=...)`

This approach would greatly simplify, streamline, and modularize the code, reduce docstring redundancies and errors, and help future maintainers see exactly where/how each layer is being computed.

src/proteus/dswx_hls.py
```diff
@@ -4377,8 +4444,10 @@ class in the browse image. False to display PSW-Agg.
                           flag_collapse_wtr_classes=FLAG_COLLAPSE_WTR_CLASSES,
                           output_files_list=build_vrt_list)
 
-    cloud, wtr_layer = _compute_and_apply_cloud_layer(wtr_2_layer, fmask,
+    cloud, wtr_layer = _compute_and_apply_cloud_layer(
+        wtr_2_layer, preliminary_cloud_layer, fmask,
         mask_adjacent_to_cloud_mode)
+    del preliminary_cloud_layer
```
nemo794, line 4450: Please see previous comment about how to better modularize the code.

src/proteus/dswx_hls.py
```diff
@@ -3479,7 +3479,7 @@ def _populate_dswx_metadata_processing_parameters(
 
     # Copernicus class for forest masking
     dswx_metadata_dict['FOREST_MASK_LANDCOVER_CLASSES'] = \
-        ', '.join(forest_mask_landcover_classes)
+        ','.join([str(c) for c in forest_mask_landcover_classes])
```
nemo794, line 3522: For the metadata, should class 201 (evergreen) also be included in this list? Even though this class is hardcoded into the algorithm, it is one of the Copernius forest classes which is being masked.

src/proteus/dswx_hls.py
```diff
@@ -3479,7 +3479,7 @@ def _populate_dswx_metadata_processing_parameters(
...
 
     # ocean masking distance from shoreline in km
     dswx_metadata_dict['OCEAN_MASKING_SHORELINE_DISTANCE_KM'] = \
```
nemo794, line 3525: Should this only be a number if a shoreline shape file is provided?

If there's no shoreline file, then we do not create an ocean mask, correct? In which case, should this be marked with an e.g. `-`, similar to how `MIN_SLOPE_ANGLE` is handled depending on `SHADOW_MASKING_ALGORITHM` used?

src/proteus/dswx_hls.py
```diff
@@ -3432,6 +3434,58 @@ def _populate_dswx_metadata_datasets(dswx_metadata_dict,
...
+        dswx_metadata_dict['MIN_SLOPE_ANGLE'] = '-'
+        dswx_metadata_dict['MAX_SUN_LOCAL_INC_ANGLE'] = '-'
+
+    # mask adjacent to cloud/cloud shadow mode
+    dswx_metadata_dict['MASK_ADJACENT_TO_CLOUD_MODE'] = \
+        mask_adjacent_to_cloud_mode
+
+    # Copernicus class for forest masking
+    dswx_metadata_dict['FOREST_MASK_LANDCOVER_CLASSES'] = \
+        ', '.join(forest_mask_landcover_classes)
```
nemo794: I left this comment elsewhere, but shouldn't class 201 (evergreen) also be included in this list for the metadata?

src/proteus/schemas/dswx_hls.yaml
```diff
@@ -79,8 +79,10 @@ runconfig:
             # Define how areas adjacent to cloud/cloud-shadow should be handled
             mask_adjacent_to_cloud_mode: enum('mask', 'ignore', 'cover', required=False)
 
-            # Copernicus CGLS Land Cover 100m forest classes
-            copernicus_forest_classes: list(int(), min=0, required=False)
+            # Copernicus CGLS Land Cover 100m forest classes to mask out from
```
nemo794, line 82: If this is a correct statements, please add a note here (and everywhere else that it is relevant) that class 201 (evergreen) will always be masked out. Thanks!

src/proteus/dswx_hls.py
```diff
@@ -4150,6 +4255,53 @@ class in the browse image. False to display PSW-Agg.
...
+
+    # create variables to compute spatial and cloud coverage
+    n_valid = np.sum(valid_array)
+    n_cloud_and_valid = np.sum(((preliminary_cloud_layer) != 0) & (valid_array))
+    del valid_array
+    total_number_of_pixels = length * width
+
+    # use floor integer to compute spatial and cloud coverage
+    spatial_coverage = int(100 * float(n_valid) / total_number_of_pixels)
+    cloud_coverage = int(100 * float(n_cloud_and_valid) / total_number_of_pixels)
```
nemo794, line 4295: Is % cloud coverage out of the entire 3660x3660 image? Or % coverage of the valid pixels?

This is perhaps a larger conversation with the group, but I'm starting to have reservations about these new metrics.

For one of the San Francisco Bay granules, here are the numbers:

```
  INPUT_HLS_PRODUCT_CLOUD_COVERAGE=25
  INPUT_HLS_PRODUCT_SPATIAL_COVERAGE=99
  CLOUD_COVERAGE=17
  SPATIAL_COVERAGE=52
```
I'm still trying to put my finger on why this is unsettling, but putting on my user hat, a couple of things stand out right off the bat:
1) Where did the clouds go?
2) Why did we lose half of the granule? To me, this has been a fantastic granule to use as a test case. But when I run queries to STAC-CMR, as a starting point I filter out anything with spatial coverage under 70%, which means I never would have found this granule.

Here's the browse image, for reference:
OPERA_DSWx-HLS_L8_30_T10SEG_20220119T184616Z_20230130T203109Z_v1 0 0_BROWSE

What are your thoughts on this? Is it still the right thing to do?

nemo794 (changes requested):
Hi @gshiroma , Thank you for updating the version number and adding the new metadata.

As you'll see in the comments, I'm a bit concerned about a couple of the new metrics, and was quite confused by splitting the computation of the cloud layer into two parts. The first might require a larger conversation with the group, and the second I provided a suggestion for how to streamline the code a bit better.

After an initial look at the outputs/metadata, everything seems ok, but I'll take a deeper look after these changes are implemented in case they introduce new bugs. Thanks!

src/proteus/dswx_hls.py
```diff
@@ -1709,14 +1706,10 @@ def _compute_and_apply_cloud_layer(wtr_2_layer, fmask,
 
        Returns
        -------
-       cloud_layer : numpy.ndarray
-              Cloud mask
-       wtr_layer : numpy.ndarray
-              Cloud-masked interpreted water layer
+       preliminary_cloud_layer : numpy.ndarray
+              Preliminary cloud mask (without the snow/ice class)
```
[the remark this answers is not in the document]
gshiroma: Thank you, @nemo794 . The output is really `preliminary_cloud_layer`, meaning preliminary CLOUD layer (not WTR). I'll update the docstrings to make it more clear.

src/proteus/dswx_hls.py
```diff
@@ -1733,8 +1726,8 @@ def _compute_and_apply_cloud_layer(wtr_2_layer, fmask,
           10 - Moderate aerosol
           11 - High aerosol
 
-    (*1) set output as 9
-    (*2) set output as 8
+    (*1) set output as WTR_CLOUD_MASKED
+    (*2) set output as WTR_SNOW_MASKED
```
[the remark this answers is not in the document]
gshiroma: I'll update the docstrings. Thank you, @nemo794

src/proteus/dswx_hls.py
```diff
@@ -1709,14 +1706,10 @@ def _compute_and_apply_cloud_layer(wtr_2_layer, fmask,
```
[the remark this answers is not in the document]
gshiroma: The string value "mask" comes from the runconfig schema. I'd prefer to not change it now.

src/proteus/dswx_hls.py
```diff
@@ -1744,18 +1737,63 @@ def _compute_and_apply_cloud_layer(wtr_2_layer, fmask,
...
+       preliminary_cloud_layer[np.bitwise_and(fmask, 2**2) == 2**2] = 1
 
     # Check Fmask cloud bit (1) => bit 2
-    cloud_layer[np.bitwise_and(fmask, 2**1) == 2**1] += 4
+    preliminary_cloud_layer[np.bitwise_and(fmask, 2**1) == 2**1] += 4
+
+    return preliminary_cloud_layer
+
+
+def _compute_and_apply_cloud_layer(wtr_2_layer, cloud_layer, fmask,
```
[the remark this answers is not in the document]
gshiroma: Thank you for you suggestion. I'm trying to change the code as minimum as possible. I had to divide the original function into two parts because cloud/cloud shadow values were needed at the beginning of the execution (to populate the CLOUD_COVERAGE metadata of all DSWx-HLS layer) and the second part that needed the WTR-2 layer. I'd like to avoid adding another division to the original function.  To avoid confusion, I'm renaming `_compute_and_apply_cloud_layer()` to `_add_snow_to_cloud_layer_and_apply_cloud_masking()`.

src/proteus/dswx_hls.py
```diff
@@ -1744,18 +1737,63 @@ def _compute_and_apply_cloud_layer(wtr_2_layer, fmask,
...
+       ----------
+       wtr_2_layer: numpy.ndarray
+              Cloud-unmasked interpreted water layer
+       cloud_layer : numpy.ndarray
+              Preliminary cloud mask (without the snow/ice class)
+       fmask: numpy ndarray
+              HLS Fmask
+       mask_adjacent_to_cloud_mode: str
+              Define how areas adjacent to cloud/cloud-shadow should be handled.
+              Options: "mask", "ignore", and "cover"
```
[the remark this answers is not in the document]
gshiroma: I'm updating the docstrings. "cover" comes from the runconfig schema and I'd prefer to avoid changing it now.

src/proteus/dswx_hls.py
```diff
@@ -4193,17 +4232,11 @@ class in the browse image. False to display PSW-Agg.
     length = image_dict['length']
     width = image_dict['width']
     invalid_ind = np.where(image_dict['invalid_ind_array'])
+    valid_array = ~image_dict['invalid_ind_array']
```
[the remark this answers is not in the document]
gshiroma: I'm keeping the unpacking of the `image_dict()` at one place. I'm also adding a line to remove that dictionary when it's no longer needed: `del image_dict` after the line you pointed out above.

src/proteus/dswx_hls.py
gshiroma: Thank you, @nemo794 . The` mask_adjacent_to_cloud_mode`  `cover` requires the WTR-2 layer to update the CLOUD layer with snow/ice values. I update the docstrings and answered some of your comments above. If it's still unclear, we can meet tomorrow (Tuesday Jan 31st) to solve this issue.

src/proteus/dswx_hls.py
gshiroma: Uhm, `FOREST_MASK_LANDCOVER_CLASSES` has nothing to do with ocean masking.

src/proteus/dswx_hls.py
gshiroma: Good point, @nemo794 . You are right, the HLS product defines the `CLOUD_COVERAGE` over valid pixels. I'm updating the code to do the same.

src/proteus/dswx_hls.py
gshiroma: > Hi @gshiroma , Thank you for updating the version number and adding the new metadata.
> 
> As you'll see in the comments, I'm a bit concerned about a couple of the new metrics, and was quite confused by splitting the computation of the cloud layer into two parts. The first might require a larger conversation with the group, and the second I provided a suggestion for how to streamline the code a bit better.
> 
> After an initial look at the outputs/metadata, everything seems ok, but I'll take a deeper look after these changes are implemented in case they introduce new bugs. Thanks!

Thank you so much for your quick review, @nemo794 ! I answered/addressed all your comments. The only remaining comments are related to the evergreen class 201 that I forward the question to John. Please, let me know if you have any other comments or concerns.

src/proteus/dswx_hls.py
nemo794: Ah - my apologies - I highlighted the incorrect lines.

Currently, for the product metadata, the block of code on lines 3484-3485 will always set `OCEAN_MASKING_SHORELINE_DISTANCE_KM` to `ocean_masking_shoreline_distance_km`. But, what happens if there is no shoreline shape file provided? Is it correct to still set the metadata field `OCEAN_MASKING_SHORELINE_DISTANCE_KM` to the default number, e.g. `1` for 1 km?

Or, if there is no shoreline shapefile provided, would it be better to set `OCEAN_MASKING_SHORELINE_DISTANCE_KM` to `-`? This approach would be similar to how the code is handling `SHADOW_MASKING_ALGORITHM` when the alternate algorithm is used ([link](https://github.com/gshiroma/PROTEUS/blob/32d1a1315b4ecb2886267f013d1ccea1d956c412/src/proteus/dswx_hls.py#L3527-L3528)).

gshiroma: Sure, sounds good, @nemo794 . I updated the code to assign `-` to the metadata field `OCEAN_MASKING_SHORELINE_DISTANCE_KM` if the shoreline shapefiile is not provided.

src/proteus/dswx_hls.py
```diff
@@ -4217,7 +4391,7 @@ class in the browse image. False to display PSW-Agg.
         landcover_mask = create_landcover_mask(
```
nemo794, line 4391: I think there's an issue with how the no data value is being set (and colored) for the LAND layer.

Running `gdalinfo` on the LAND.tif layer, the normal `NoData Value=...` field is missing, and the colortable entry assigned for value 255 is `255: 255,255,255,255`. (The other layers have this set as `255: 0,0,0,0`).

I'm not sure what's happening; could you please look into this? The Product Spec document says that there should be a fill value for this layer.

src/proteus/dswx_hls.py
```diff
@@ -4156,6 +4283,53 @@ class in the browse image. False to display PSW-Agg.
...
+    cloud_coverage = int(100 * float(n_cloud_and_valid) / n_valid)
+
+    # print spatial and cloud coverage
+    logger.info(f'    spatial coverage [%]:  {spatial_coverage}')
+    logger.info(f'    cloud coverage [%]:  {cloud_coverage}')
+
+    # update DSWx-HLS metadata dictionary
+    dswx_metadata_dict['SPATIAL_COVERAGE'] = spatial_coverage
+    dswx_metadata_dict['CLOUD_COVERAGE'] = cloud_coverage
+
```
nemo794, line 4332: Git is not allowing me to comment on the relevant lines, but could you please confirm that it is correct that there is no fill data value for the SHAD.tif layer?

The Product Spec is not specific about this, and running `gdalinfo` on one of the *SHAD.tif output files shows that there is no fill data value specified, and then in the color table 255 is fully visible (not transparent). Thanks!

gshiroma: Hi @nemo794 , thank you for checking, but I think the shadow layer doesn't have fill value. We expect the DEM to be available globally. If it's not available for the area, the `_check_ancillary_inputs()` should raise an error.

nemo794: Sounds good! And thank you for updating the Product Spec document to match. 🚀

src/proteus/dswx_hls.py
nemo794: Per meeting with J. Jones, class `201` should not be included. (This was a misunderstanding on my part.)

src/proteus/dswx_hls.py
```diff
@@ -3432,6 +3521,64 @@ def _populate_dswx_metadata_datasets(dswx_metadata_dict,
...
+    else:
+        dswx_metadata_dict['MIN_SLOPE_ANGLE'] = '-'
+        dswx_metadata_dict['MAX_SUN_LOCAL_INC_ANGLE'] = '-'
+
+    # mask adjacent to cloud/cloud shadow mode
+    dswx_metadata_dict['MASK_ADJACENT_TO_CLOUD_MODE'] = \
+        mask_adjacent_to_cloud_mode
+
+    # Copernicus class for forest masking
+    dswx_metadata_dict['FOREST_MASK_LANDCOVER_CLASSES'] = \
```
nemo794, line 3571: If `forest_mask_landcover_classes` is set to `null` (or an empty list) in the runconfig, it looks like this metadata field would be empty. Should it be set to a dash `-` instead?

gshiroma: Sounds good, @nemo794 . Just updated "-" rather than "" to empty `FOREST_MASK_LANDCOVER_CLASSES`. Thanks

src/proteus/dswx_hls.py
```diff
@@ -1693,49 +1694,51 @@ def _compute_diagnostic_tests(blue, green, red, nir, swir1, swir2,
...
+            _add_snow_to_cloud_layer_and_apply_cloud_masking())
 
        Returns
        -------
-       cloud_layer : numpy.ndarray
-              Cloud mask
-       wtr_layer : numpy.ndarray
-              Cloud-masked interpreted water layer
+       preliminary_cloud_layer : numpy.ndarray
+            Preliminary cloud mask (without the snow/ice class)
```
nemo794, line 1719: Could the docstring please be updated to something like:

```suggestion
            Preliminary cloud mask (without the snow/ice class)
            with dtype of uint8 and values assigned as follows:
                0: Not masked
                1: Cloud shadow or adjacent to cloud/cloud shadow
                4: Cloud
                5: Cloud and class 1 (cloud shadow or adjacent to cloud/cloud shadow)
                255: Fill value (no data)

       See Also
       ---------
       _add_snow_to_cloud_layer : Add the snow bit encodings from `Fmask` to `preliminary_cloud_layer`
```
numpy docstring reference: [link](https://numpydoc.readthedocs.io/en/latest/format.html#see-also)

Note: the values I wrote above might be incorrect. But, this is why the docstring will be very helpful. Thanks!

src/proteus/dswx_hls.py
```diff
@@ -1744,21 +1747,66 @@ def _compute_and_apply_cloud_layer(wtr_2_layer, fmask,
...
+       to the CLOUD layer.
+       This function succeeds the function _compute_preliminary_cloud_layer()
+       and preceeds _apply_cloud_masking()
+
+       Parameters
+       ----------
+       wtr_2_layer: numpy.ndarray
+            Cloud-unmasked interpreted water layer
+       cloud_layer : numpy.ndarray
+            Preliminary cloud mask (without the snow/ice class)
```
nemo794, line 1776: Something like:

```suggestion
            Preliminary cloud mask (without the snow/ice class)
            with dtype of uint8 and values assigned as follows:
                0: Not masked
                1: Cloud shadow or adjacent to cloud/cloud shadow
                4: Cloud
                5: Cloud and class 1 (cloud shadow or adjacent to cloud/cloud shadow)
                255: Fill value (no data)
```

src/proteus/dswx_hls.py
gshiroma: Just update the docstrings. Thank you, @nemo794

src/proteus/dswx_hls.py
gshiroma: Just update the docstrings. Thank you, @nemo794

src/proteus/dswx_hls.py
```diff
@@ -1744,21 +1747,66 @@ def _compute_and_apply_cloud_layer(wtr_2_layer, fmask,
...
+            Define how areas adjacent to cloud/cloud-shadow should be handled.
+            Options: "mask" - mask out these areas marking them as cloud
+            shadow (default); "ignore" - ignore the adjacent to cloud/
+            cloud shadow classification; and 'cover' - covers these areas
+            with a dilation algorithm
+
+       Returns
+       -------
+       cloud_layer : numpy.ndarray
+              Cloud mask
```
nemo794, line 1795:
```suggestion
              Cloud mask. This is the finished DSWx-HLS `CLOUD` layer,
              with dtype of uint8 and values assigned as follows:
                  0: Not masked
                  1: Cloud shadow or adjacent to cloud/cloud shadow
                  2: Snow/ice
                  3: Snow/ice and class 1 (cloud shadow or adjacent to cloud/cloud shadow)
                  4: Cloud
                  5: Cloud and class 1 (cloud shadow or adjacent to cloud/cloud shadow)
                  6: Cloud and snow/ice
                  7: Cloud, snow/ice, and class 1 (cloud shadow or adjacent to cloud/cloud shadow)
                  255: Fill value (no data)

See Also
---------
`_compute_preliminary_cloud_layer` : Can use to create input `cloud_layer`
```

src/proteus/dswx_hls.py
nemo794: Thank you for updating `CLOUD COVERAGE`.

Regarding `SPATIAL_COVERAGE`, I'll just note here that as discussed in yesterday's meeting, the decision is to keep `SPATIAL_COVERAGE` computed as-is, and then add a new metadata field that computes spatial coverage over pixels that are not marked as "ocean" by the ocean mask.

src/proteus/dswx_hls.py
```diff
@@ -1744,21 +1747,66 @@ def _compute_and_apply_cloud_layer(wtr_2_layer, fmask,
...
-    cloud_layer[np.bitwise_and(fmask, 2**1) == 2**1] += 4
+    preliminary_cloud_layer[np.bitwise_and(fmask, 2**1) == 2**1] += 4
 
-    # If cloud (1) or cloud shadow (3), mark WTR as WTR_CLOUD_MASKED
-    wtr_layer[cloud_layer != 0] = WTR_CLOUD_MASKED
+    return preliminary_cloud_layer
+
+
+
+def _add_snow_to_cloud_layer(wtr_2_layer, cloud_layer, fmask,
```
nemo794, line 1771: Currently, this docstring says that the function takes in an argument named `cloud_layer`, and then returns an argument named `cloud_layer`. It's not great practice to use the same variable name for inputs and outputs. Plus, in the code itself, it looks like the input `cloud_layer` is being modified directly and so there's no need for it to be returned.

Two suggestions for how to approach this:
1) change the input variable name to be `preliminary_cloud_layer`, then make a copy of it in the code and return `cloud_layer`. This is the style of other functions in the codebase, and much clearer about what the input should be.
2) update the docstring to say that the input cloud layer is modified by the function, make sure to detail _how_ the function is modifying the actual values of the input layer, and also remove the `Return` statement. This strategy is less preferable to me, as it is contrary to the coding style used elsewhere in the code.

gshiroma: I prefer to avoid making an unnecessary copies of the array. I think the code "as is" is easier to follow.

src/proteus/dswx_hls.py
nemo794: Thank you for filling out the docstring. I'm not sure if something got mangled in the git merge, but now the output variable name and type disappeared, the indentation is off, and there should be an extra newline between the `Return` block and the `See Also` block. Could you please look into this? Thank you!

nemo794: Your recent commit just fixed the spacing in the `_add_snow_to_cloud_layer` docstring (thank you!), but the spacing issues mentioned are still present in the `_compute_preliminary_cloud_layer` docstring. Could you please update? Thank you!

src/proteus/dswx_hls.py
nemo794: In that case, could you please follow suggestion 2?

gshiroma: I think the way it is now it's easier to understand. `cloud_layer` is mutable, but by returning it, we are making explicit that that's the output of the function. The input `preliminary_cloud_layer` and output `cloud_layer` are the same array, but I think the code this way is easier to follow. I updated the docstrings to make this point more clear:
```       cloud_layer : numpy.ndarray
            Preliminary cloud mask (without the snow/ice class). This array
            is updated within the function (mutable numpy.ndarray) and returned
            after the snow/ice class is added to the layer.```

Comment (nemo794) [reply src/proteus/dswx_hls.py:1773]: I agree that it's better to be explicit, both for the input arguments and the output arguments. In that case, please go with suggestion 1. There are so many other places in the code where extra copies of arrays are being created, I don't think that one additional occurance of that is going to matter.

Comment (gshiroma) [reply src/proteus/dswx_hls.py:1773]: I don't see a reason to add an unnecessary copy of data. The other array copies in the code are needed to create different layers. If that's not the case, please let me know because then we should fix that too.

Comment (nemo794) [reply src/proteus/dswx_hls.py:1773]: Is there a third approach? The requirements for the new approach would be that this function can not both mutate an input array and return that same array. It must only do one of these things. This is significantly cleaner.

Some pros and cons of each option: Mutating an input array makes for code that is very hard to debug and understand (as an example, mutating input arguments was a significant reason why the original QA code was nearly impossible to understand), but it avoids copying an array. Copying the array is much cleaner for the user to understand and is one of the foundations of functional programming, but does create a (temporary) copy of the array.

Comment (gshiroma) [reply src/proteus/dswx_hls.py:1773]: I guess the third approach is what we are doing lol. The behavior is described in the docstrings and also, this function intended for internal use (startswith `_`), so it won't be used by other modules. It's also very specific to a part of the code (add snow class to the cloud layer). Finally, having `preliminary_cloud_layer` and `cloud_layer` make it clear what is preliminary and final.

File: src/proteus/dswx_hls.py [hunk 21]
```diff
@@ -1169,73 +1176,76 @@ def _apply_aerosol_class_remapping_wtr1_class(wtr_1_layer,
...
     for fmask_value in fmask_values:
-        if to_mask_array is None:
-            to_mask_array = fmask == fmask_value
+        if to_remap_array is None:
+            to_remap_array = fmask == fmask_value
         else:
-            to_mask_array |= fmask == fmask_value
-    to_mask_array &= wtr_1_layer == input_wtr1_class
-    wtr_1_layer[to_mask_array] = output_wtr1_class
+            to_remap_array |= fmask == fmask_value
```
nemo794, line 1184: This can probably be simplified with `numpy.isin` ([link](https://numpy.org/doc/stable/reference/generated/numpy.isin.html#numpy.isin)).

gshiroma: Thank you, @nemo794 ! I didn't know that function, it definitely simplifies the logic

nemo794 (commented):
Once again, several great new features in this PR, including some complicated but vital improvements to the algorithm and code. I have several concerns about the coding style in terms of long term readability and maintainability, but the code seems to work.

There are many tweaks to the algorithm are still occurring offline, but I'll push these comments now. Thanks!

src/proteus/dswx_hls.py
```diff
@@ -1904,26 +1944,18 @@ def _add_snow_to_cloud_layer(wtr_2_layer, cloud_layer, fmask,
...
-                shadow)
-            6: Cloud and snow/ice
-            7: Cloud, snow/ice, and class 1 (cloud shadow or adjacent to
-                cloud/cloud shadow)
+            values assigned following the bit encoding:
+
+            0: Cloud shadow or adjacent to cloud/cloud shadow
+            1: Snow/ice
+            2: Cloud
+            3: Class reassignment due to aerosol interpolation errors (3)
```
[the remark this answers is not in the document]
gshiroma: This is the bit encoding. I'm updating the docstrings to make it more clear.

src/proteus/dswx_hls.py
```diff
@@ -3677,23 +3677,23 @@ def _populate_dswx_metadata_datasets(dswx_metadata_dict,
         dswx_metadata_dict['DEM_SOURCE'] = \
             os.path.basename(dem_file)
     else:
-        dswx_metadata_dict['DEM_SOURCE'] = '(not provided)'
+        dswx_metadata_dict['DEM_SOURCE'] = '-'
```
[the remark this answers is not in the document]
gshiroma: Yeah, the problem is that, for example, the shoreline shapefile can be provided but not used, that's why I changed it from `not provided` to `-` (which BTW was already used for some other metadata fields). Any suggestions?

src/proteus/dswx_hls.py
```diff
@@ -1713,17 +1713,19 @@ def _compute_preliminary_cloud_layer(fmask, mask_adjacent_to_cloud_mode):
 
        Returns
        -------
-            Preliminary cloud mask (without the snow/ice class)
-            with dtype of uint8 and values assigned as follows:
+            preliminary_cloud_layer : numpy.ndarray
+                Preliminary cloud mask (without the snow/ice class)
+                with dtype of uint8 and values assigned as follows:
```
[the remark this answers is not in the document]
gshiroma: Sorry, I'm having a hard time figuring out what is wrong here. The indentation matches "Returns". Is it possible that this comment is from a previous version of the code?

src/proteus/dswx_hls.py
```diff
@@ -1792,8 +1792,21 @@ def _add_snow_to_cloud_layer(wtr_2_layer, cloud_layer, fmask,
...
+        2: Snow/ice
+        3: Snow/ice and class 1 (cloud shadow or adjacent to cloud/cloud
+        shadow)
+        4: Cloud
+        5: Cloud and class 1 (cloud shadow or adjacent to cloud/cloud
+        shadow)
+        6: Cloud and snow/ice
+        7: Cloud, snow/ice, and class 1 (cloud shadow or adjacent to
+        cloud/cloud shadow)
+        255: Fill value (no data)
```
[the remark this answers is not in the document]
gshiroma: I'm not finding these lines. Is it possible that this comment is from a previous version of the code?

src/proteus/dswx_hls.py
```diff
@@ -1781,7 +1781,9 @@ def _add_snow_to_cloud_layer(wtr_2_layer, cloud_layer, fmask,
        wtr_2_layer: numpy.ndarray
             Cloud-unmasked interpreted water layer
        cloud_layer : numpy.ndarray
-            Preliminary cloud mask (without the snow/ice class)
+            Preliminary cloud mask (without the snow/ice class). This array
+            is updated within the function (mutable numpy.ndarray) and returned
+            after the snow/ice class is added to the layer.
```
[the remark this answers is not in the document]
gshiroma: As discussed in the other thread, copying the array without necessity is out of discussion. I also think it's clearer to have `preliminary_cloud_layer` separated from the `cloud_layer`. The reader can see immediately what's preliminary and what's not. I can change it to have this PR merged, but I still prefer the current implementation compared to the two options that were presented.

src/proteus/dswx_hls.py
```diff
@@ -4544,7 +4544,6 @@ class in the browse image. False to display PSW-Agg.
     cloud = _add_snow_to_cloud_layer(
         wtr_2_layer, preliminary_cloud_layer, fmask,
         mask_adjacent_to_cloud_mode)
-    del preliminary_cloud_layer
```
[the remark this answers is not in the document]
gshiroma: `preliminary_cloud_layer` and `cloud_layer` are the same array, so the former doesn't take any extra memory.

gshiroma:
> Once again, several great new features in this PR, including some complicated but vital improvements to the algorithm and code. I have several concerns about the coding style in terms of long term readability and maintainability, but the code seems to work.
> 
> There are many tweaks to the algorithm are still occurring offline, but I'll push these comments now. Thanks!

Thank you, @nemo794, for your review. I addressed your comments above. Please, let me know if you still have any other comments or concerns.

nemo794 (approved):
Code works! Will leave it to you to decide how to handle any of the remaining comments. Thanks again!

gshiroma:
> 

Thank you, @nemo794 !

hfattahi (approved):
Thanks @gshiroma. Based on the sample products produced with this PR, I think we can move forward and merge these changes.

gshiroma:
Thank you, @hfattahi

PR #39, opened 2023-02-18
Fix and improve ancillary inputs checks.

gshiroma:
This PR fixes and improves the ancillary inputs checks.

The current implementation expects the output of the GR function `TransformPoint()` to be (x, y, z). However, for geographic coordinate system (e.g., EPSG 4326),  the output is provided as (y, x, z) which is causing the ancillary input check to fail. This PR fixes this issue by setting `osr.OAMS_TRADITIONAL_GIS_ORDER` to the output SRS `AxisMappingStrategy` in order to force the transformation output to be in the form (x, y, z). More information can be found [here](https://gdal.org/tutorials/osr_api_tut.html#crs-and-axis-order).

This function also fixes/improves the test that the tile polygon is within the input file polygon. Previously, this was done point-by-point and required that the ancillary input and the tile intersected, which is not correct. This PR fixes the checks.

Finally, this PR also removes unnecessary code that was checking if `projection.upper() == 'WGS84'`. This check is completely unnecessary and the function `ImportFromProj4(projection)` accounts for the different types of projection that the input may have.

### DIFF
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -2944,7 +2944,9 @@ def _get_tile_srs_bbox(tile_min_y_utm, tile_max_y_utm,
         try:
             polygon_srs.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
         except AttributeError:
-            pass
+            logger.warning('WARNING Could not set the ancillary input SRS axis'
+                           ' mapping strategy (SetAxisMappingStrategy())'
+                           ' to osr.OAMS_TRADITIONAL_GIS_ORDER')
     transformation = osr.CoordinateTransformation(tile_srs, polygon_srs)
 
     elevation = 0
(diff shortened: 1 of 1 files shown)


src/proteus/dswx_hls.py
```diff
@@ -2939,6 +2938,13 @@ def _get_tile_srs_bbox(tile_min_y_utm, tile_max_y_utm,
               Tile maximum X-coordinate (polygon SRS)
     """
 
+    # forces returned values from TransformPoint() to be (x, y, z)
+    # rather than (y, x, z) for geographic SRS
+    if polygon_srs.IsGeographic():
+        try:
+            polygon_srs.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
+        except AttributeError:
+            pass
```
collinss-jpl, line 2947: If you happen to hit this `except`, I would think you'd want to at least log that it occurred.

collinss-jpl (approved):
Fixes look good to me, looking forward to getting the release

collinss-jpl (approved):

hfattahi (approved):
Thank you @gshiroma . The logic is clear. LGTM.

src/proteus/dswx_hls.py
gshiroma: Thank you, @collinss-jpl ! Just updated the code to log it as a warning message.

gshiroma:
Thank you so much for you review, @collinss-jpl and @hfattahi . I'm merging this PR now.

PR #44, opened 2023-03-06
Quick updates to the LAND layer

gshiroma:
This PR brings two quick updates to the LAND layer:
- Add the Copernicus Land Cover "urban/built-up" class 50, permanent water bodies class 80, and open see class 200 to the runconfig field `forest_mask_landcover_classes`;
- Fix the year of the LAND developed classes to be assigned from the WorldCover metadata (average of `time_start` and `time_end`).

### DIFF
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -1053,38 +1053,43 @@ def create_landcover_mask(copernicus_landcover_file,
         # the Worldcover dataset year is extracted from the average date times
         worldcover_time_range = (worldcover_time_end - worldcover_time_start)
         year = (worldcover_time_start + worldcover_time_range / 2.0).year
-        logger.info('ESA WorldCover map year: {year}'
+        logger.info('    ESA WorldCover map year: {year}'
                     '(source: WorldCover file metadata)')
 
-    else:
+    elif worldcover_file_description:
         logger.warning('WARNING Could not read the ESA WorldCover 10m metadata'
                        ' fields `time_start` and/or `time_end`')
         year_start = 2000
         year_end_next = 2100
-        for year in range(year_start, year_end_next)
+        for year in range(year_start, year_end_next):
             if str(year) in worldcover_file_description:
-                logger.info('ESA WorldCover map year: {year}'
+                logger.info('    ESA WorldCover map year: {year}'
                             '(source: WorldCover file description)')
                 break
         else:
             year = 2000
             logger.warning('WARNING Could not infer the ESA WorldCover 10m'
                            ' data year from the WorldCover file description.'
-                           ' Considering it as 0 (2000)')
+                           ' Considering year as 2000')
+    else:
+        year = 2000
+        logger.warning('WARNING Could not read the ESA WorldCover 10m metadata'
+                       ' fields `time_start` and/or `time_end`'
+                       ' Considering year as 2000')
 
-    year -= 2000
+    year_offset = year - 2000
 
     # majority of pixels are urban
     low_intensity_developed_class = \
         (dswx_hls_landcover_classes_dict['low_intensity_developed_offset'] +
-         year)
+         year_offset)
     _update_landcover_array(hierarchy_combined, urban_aggregate_sum,
                             threshold_list[1], low_intensity_developed_class)
 
     # high density urban at 7/9 or higher
     high_intensity_developed_class = \
         (dswx_hls_landcover_classes_dict['high_intensity_developed_offset'] +
-         year)
+         year_offset)
     _update_landcover_array(hierarchy_combined, urban_aggregate_sum,
                             threshold_list[2], high_intensity_developed_class)
 
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -894,7 +894,8 @@ def _update_landcover_array(conglomerate_array, agg_sum, threshold,
 
 
 def create_landcover_mask(copernicus_landcover_file,
-                          worldcover_file, output_file, scratch_dir,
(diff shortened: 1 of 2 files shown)


hfattahi (approved):
The changes affecting the year LGTM. The added land cover class of 50 should be confirmed. But the implementation looks fine to me.

src/proteus/dswx_hls.py
```diff
@@ -996,8 +996,21 @@ def create_landcover_mask(copernicus_landcover_file,
...
+    del worldcover_gdal_ds
+    worldcover_datetime_format = '%Y-%m-%dT%H:%M:%SZ'
+    worldcover_time_start = datetime.strptime(worldcover_metadata['time_start'],
+                                              worldcover_datetime_format)
+    worldcover_time_end = datetime.strptime(worldcover_metadata['time_end'],
+                                            worldcover_datetime_format)
+
+    # the Worldcover dataset year is extracted from the average date times
+    worldcover_time_range = (worldcover_time_end - worldcover_time_start)
+    year = (worldcover_time_start + worldcover_time_range / 2.0).year - 2000
```
nemo794, line 1011: Looking at the WorldCover user guides ([2020](https://esa-worldcover.s3.eu-central-1.amazonaws.com/v100/2020/docs/WorldCover_PUM_V1.0.pdf), [2021](https://esa-worldcover.s3.eu-central-1.amazonaws.com/v200/2021/docs/WorldCover_PUM_V2.0.pdf)), it looks like each WorldCover map only contains data from a single year. What scenario are you thinking of where the data would be from multiple years?

Otherwise, let's simplify this code.

gshiroma: This is an example that I have here. From the WorldCover metadata:
```
 time_end=2020-12-31T23:59:59Z
 time_start=2020-01-01T00:00:00Z
```
I'm just taking the center (average) date because these two dates are in the "edge" of the year. The code will be robust to any changes that they apply on their side.

nemo794: Got it. Why not take the earlier of the two years, instead of the average? Per the DSWx-HLS product spec, the year in the LAND layer denotes "...first year development is present".

gshiroma: For example, if they changed it to:
``` 
 time_end=2021-01-01T00:00:00Z
 time_start=2020-01-01T00:00:00Z
```
or to
``` 
 time_end=2020-12-31T23:59:59Z
 time_start=2019-12-31T23:59:59Z
```
I know these changes are unlikely though

hfattahi: The average makes sense to me. It covers the normal cases like the existing one as well unlikely cases.

gshiroma: This code refers to the WorldCover dataset that was given to the SAS. We want to know if this WorldCover dataset refers to year 2020 or 2021. For that, taking the center date is more robust.
The product specs refers to something else, it refers to the updating the LAND layer incrementally.

gshiroma: John's initial idea was to have a "semi-static" global LAND layer. So, for example, OPERA would generate the global LAND layer for 2022. Then, for 2023, it would take the LAND from 2022 and update with the Copernicus Landcover 2023 and ESA Worldcover 2023 and generate the OPERA LAND 2023. If a pixel in the LAND layer 2022 was marked as urban/builtup 2022, and it's still urban/builtup in 2023, we'd keep the first time that it was marked as urban/builtup (in this example, 2022).

gshiroma: I actually asked John today if we should keep with that classification considering that we don't have this "semi-static" global LAND layer anymore and we do write in the DSWx-HLS a description of the land cover inputs. He said he thinks it's useful to users to have the current urban/built up date (even though we are not keeping historical values anymore). So that date will always point to the WorldCover dataset that was provided to the SAS

nemo794 (approved):
The CI checks are not passing, but otherwise LGTM.

gshiroma:
@hfattahi and @nemo794 , as discussed, I did some changes to this PR. You guys may want to take a last look before I merge it. Thanks.

gshiroma:
> Thanks @gshiroma looks good to me.

Thank you again, @hfattahi .

src/proteus/dswx_hls.py
```diff
@@ -1038,15 +1041,38 @@ def create_landcover_mask(copernicus_landcover_file,
...
+                logger.info('ESA WorldCover map year: {year}'
+                            '(source: WorldCover file description)')
+                break
+        else:
+            year = 2000
+            logger.warning('WARNING Could not infer the ESA WorldCover 10m'
+                           ' data year from the WorldCover file description.'
+                           ' Considering it as 0 (2000)')
+
+    year -= 2000
```
nemo794: Re: year -- Documenting here a summary our offline discussion from yesterday:

After this release, it would be best to remove all references to the year in the LAND data values themselves, and instead simply use the metadata field `WORLDCOVER_SOURCE` to document the year of the WorldCover input source for the users.

Based on that change, it would be really nice to update the data value classifications to something simpler, such as:

```
0: [Everything else]
1: Water, Wetland, Mangrove Forest
2: Low Intensity Developed
3: High Intensity Developed
4: A variety of  Forest Classes
255: Fill value (no data)
```

nemo794:
This branch looks good and is running on test datasets again. Nice! LGTM.

Quick reminder that all of the `*_file_description` fields in the runconfig must be populated.

gshiroma:
> This branch looks good and is running on test datasets again. Nice! LGTM.
> 
> Quick reminder that all of the `*_file_description` fields in the runconfig must be populated.

Thank you, @nemo794 !

PR #45, opened 2023-03-07
Quick PR to address pylint issues

gshiroma:
This PR addresses issues identified by pylint. Besides the style changes, there's just one bug that was caught by pylint (in line 2622). This bug does not affect the DSWx-HLS product. The variable `band_name` was being used instead of the variable `layer_name` (which is the variable used in the for loop). For the current DSWx-HLS product, all layers are single band and `band_name` is always equal to `layer_name`, so the bug was not affecting the product (and therefore not being caught), but we should fix it anyway.


nemo794 (approved):
LGTM

gshiroma:
> LGTM

Thank you, @nemo794 !

hfattahi (approved):

PR #48, opened 2023-03-09
Final changes to DSWx-HLS SAS for v1.0

gshiroma:
This PR provides final changes to the DSWx-HLS SAS for v1.0:
- Set default for `apply_ocean_masking` (in the default runconfig file) to `False`;
- Updates DSWx-HLS metadata:
    - Adds `AEROSOL_CLASS_REMAPPING_ENABLED` and `OCEAN_MASKING_ENABLED`;
    - Substitutes fields with value `-` by more meaningful values (addresses @nemo794 comment [here](https://github.com/nasa/PROTEUS/pull/38#discussion_r1127180356);
- Update URL pointing to the updated golden dataset in Zenodo (to be updated after all changes are approved).

### DIFF
--- a/src/proteus/defaults/dswx_hls.yaml
+++ b/src/proteus/defaults/dswx_hls.yaml
@@ -103,7 +103,7 @@ runconfig:
             # Copernicus CGLS Land Cover 100m forest classes to mask out from
             # the WTR-2 and WTR layer due to dark reflectance that is usually
             # misinterpreted as water.
-            forest_mask_landcover_classes: [20, 50, 80, 111, 113, 115, 116, 121, 123, 125, 126, 200]
+            forest_mask_landcover_classes: [20, 50, 111, 113, 115, 116, 121, 123, 125, 126]
 
             # Ocean masking distance from shoreline in km
             ocean_masking_shoreline_distance_km: 1
--- a/src/proteus/defaults/dswx_hls.yaml
+++ b/src/proteus/defaults/dswx_hls.yaml
@@ -67,7 +67,7 @@ runconfig:
             check_ancillary_inputs_coverage: True
 
             # Apply ocean masking
-            apply_ocean_masking: True
+            apply_ocean_masking: False
 
             # Apply aeresol class remapping
             apply_aerosol_class_remapping: True
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -1569,7 +1569,7 @@ def _get_cloud_layer_ctable():
     # Light gray - Cloud
     mask_ctable.SetColorEntry(12, (192, 192, 192))
     # Gray - Cloud and cloud shadow
-    mask_ctable.SetColorEntry(13, (175, 175, 175))
+    mask_ctable.SetColorEntry(13, (127, 127, 127))
     # Magenta - Cloud and snow/ice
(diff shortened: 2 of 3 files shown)


src/proteus/dswx_hls.py
```diff
@@ -3943,14 +3953,17 @@ def _populate_dswx_metadata_processing_parameters(
         dswx_metadata_dict['FOREST_MASK_LANDCOVER_CLASSES'] = \
             ','.join([str(c) for c in forest_mask_landcover_classes])
     else:
-        dswx_metadata_dict['FOREST_MASK_LANDCOVER_CLASSES'] = '-'
+        dswx_metadata_dict['FOREST_MASK_LANDCOVER_CLASSES'] = 'EMPTY'
+
+    # ocean masking
+    dswx_metadata_dict['OCEAN_MASKING_ENABLED'] = \
+        'TRUE' if apply_ocean_masking else 'FALSE'
```
nemo794, line 3960: Please check my understanding: If a shoreline shapefile is not provided, then ocean masking will not be applied, regardless of how `apply_ocean_masking` is set.

If so, then in that scenario, should `OCEAN_MASKING_ENABLED` also be set to `False`? Currently, it looks like the code will set it to `TRUE`?

src/proteus/defaults/dswx_hls.yaml
```diff
@@ -67,7 +67,7 @@ runconfig:
             check_ancillary_inputs_coverage: True
 
             # Apply ocean masking
-            apply_ocean_masking: True
+            apply_ocean_masking: False
```
nemo794, line 70: Nice!

src/proteus/dswx_hls.py
gshiroma: This is verified in `_check_ancillary_inputs()` if `apply_ocean_masking` is set but a shoreline shapefile is not provided, the SAS will raise an error in that function (btw, I'm can simplify the next condition `if apply_ocean_masking and shoreline_shapefile:`).

hfattahi:
The changes look good to me. I will wait for a real data test

nemo794 (approved):
Using all defaults, the code appears to work and produce the desired outputs on multiple test datasets. The additional metadata fields are included, and the `NOT_PROVIDED` entries populate correctly. Nice!

Looking ahead, it would be beneficial to discuss and make sure everyone is in agreement on the behavior for when a subset of the three primary ancillary files (Worldcover, Landcover, and DEM) are missing. (The shoreline shapefile is handled differently.) However, for Ops processing, those three ancillary files will be included, and in that nominal case the code works well.

gshiroma:
> Using all defaults, the code appears to work and produce the desired outputs on multiple test datasets. The additional metadata fields are included, and the `NOT_PROVIDED` entries populate correctly. Nice!
> 
> Looking ahead, it would be beneficial to discuss and make sure everyone is in agreement on the behavior for when a subset of the three primary ancillary files (Worldcover, Landcover, and DEM) are missing. (The shoreline shapefile is handled differently.) However, for Ops processing, those three ancillary files will be included, and in that nominal case the code works well.

Thank you, @nemo794 !

gshiroma:
@hfattahi , just letting you know that I finished the changes to this PR

hfattahi (approved):
Thank you @gshiroma !

PR #50, opened 2023-03-21
More antimeridian ("dateline") crossing fixes

gshiroma:
This PR addresses an issue reported by the PGE (thank you, @collinss-jpl) that is affecting some datasets that cross the dateline. The problem is that a buffer declared in the line [here](https://github.com/nasa/PROTEUS/blob/8b6aad1a67f2f4946aae7b5ff046de9df5b66a69/src/proteus/dswx_hls.py#L4385) that is used to check the ancillary input's coverage at the right side of the dateline is too small. In this PR, I'm increasing that buffer from 1 arcsec (0.00027777 deg ~= 30m) to 33 arcsec (about 1km).

I'm taking advantage that we are updating proteus and I'm also adding some other changes:
- Small indentation fixes, white spaces, etc;
- Add condition `min_x < -170` to detect when dateline crossing requires cropping/reprojected of the ancillary input. Dateline crossing only requires reprojection of the ancillary input if this input is in geographic coordinates and its longitude domain is represented using the longitude range [-180, +180] rather than [0, +360]. For example, if a DEM was provided using [0, +360] longitude values, a tile that crosses the dateline (~+180 deg long.) would be fully covered by the DEM without need of cropping or reprojecting it. To detect if the DEM uses the [-180, +180], we just assert that `min_x` is below `-170`.

### DIFF
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -30,6 +30,9 @@ If enabled, set all negative reflectance values to 1 before scaling is applied
 '''
 FLAG_CLIP_NEGATIVE_REFLECTANCE = True
 
+# Buffer for antimeridian crossing test (33 arcsec: ~ 1km)
+ANTIMERIDIAN_CROSSING_RIGHT_SIDE_TEST_BUFFER = 33 * 0.0002777 
+  
 LANDCOVER_LAT_MAX = 80
 LANDCOVER_LAT_MIN = -60
 WORLDCOVER_LAT_MAX = 84
@@ -4429,9 +4432,9 @@ def _check_ancillary_inputs(check_ancillary_inputs_coverage,
             logger.info(f'    left side (-180 -> +180): {check_1_str}')
 
             # Right side of the antimeridian crossing: +180 -> +360
-            buffer_in_degrees = 33 * 0.0002777  # buffer of 33 arcsec: ~ 1km
             file_polygon_2 = _get_ogr_polygon(
-                max_x + buffer_in_degrees, 90, max_x + 360, -90, file_srs)
+                max_x + ANTIMERIDIAN_CROSSING_RIGHT_SIDE_TEST_BUFFER,
+                90, max_x + 360, -90, file_srs)
             intersection_2 = tile_polygon.Intersection(file_polygon_2)
             file_polygon_2 = _get_ogr_polygon(min_x + 360, max_y,
                                               max_x + 360, min_y,
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -3155,17 +3155,21 @@ def _antimeridian_crossing_requires_special_handling(
 
     '''
 
-    # The dateline crossing by the tile is tested by the condition:
-    #     `tile_min_x < 180 and tile_max_x >= 180`
-    # The ancillary input only requires reprojection if it is in
-    # geographic coordinates and if its longitude domain is represented
-    # within the [-180, +180] range, verified by the test `min_x < -170`,
-    # rather than the [0, +360] interval. There's no specific reason why
-    # -170 is used. It could be -160, or even 0.
+    # Flag to indicate if the if the MGRS tile crosses the antimeridian.
+    flag_tile_crosses_antimeridian = tile_min_x < 180 and tile_max_x >= 180
 
+    # Flag to test if the ancillary input file is in geographic
+    # coordinates and if its longitude domain is represented
+    # within the [-180, +180] range, rather than the [0, +360] interval.
+    # This is verified by the test `min_x < -170`,
+    flag_input_geographic_and_longitude_not_0_360 = \
+        file_srs.IsGeographic() and file_min_x < -170
+
+    # If both are true, tile requires special handling due to the
+    # antimeridian crossing
     flag_requires_special_handling = (
-        file_srs.IsGeographic() and file_min_x < -170 and
-        tile_min_x < 180 and tile_max_x >= 180)
+        flag_tile_crosses_antimeridian and
+        flag_input_geographic_and_longitude_not_0_360)
 
     return flag_requires_special_handling
 
--- a/src/proteus/dswx_hls.py
+++ b/src/proteus/dswx_hls.py
@@ -3168,7 +3168,8 @@ def _antimeridian_crossing_requires_special_handling(
     # Flag to test if the ancillary input file is in geographic
     # coordinates and if its longitude domain is represented
     # within the [-180, +180] range, rather than the [0, +360] interval.
-    # This is verified by the test `min_x < -170`,
+    # This is verified by the test `min_x < -170`. There's no specific reason
+    # why -170 is used. It could be -160, or even 0.
     flag_input_geographic_and_longitude_not_0_360 = \
         file_srs.IsGeographic() and file_min_x < -170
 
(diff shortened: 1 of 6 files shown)


src/proteus/dswx_hls.py
```diff
@@ -3224,11 +3233,17 @@ def _warp(input_file, geotransform, projection,
...
+    #     `tile_min_x < 180 and tile_max_x >= 180`
+    # The ancillary input only requires reprojection if it is in
+    # geographic coordinates and if its longitude domain is represented
+    # within the [-180, +180] range, verified by the test `min_x < -170`,
+    # rather than the [0, +360] interval.
+    if not (file_srs.IsGeographic() and file_min_x < -170 and
+            tile_min_x < 180 and tile_max_x >= 180):
+
+        # if it doesn't cross the antimeridian, reproject
+        # input file using gdalwarp
```
nemo794, line 3246: Perhaps this could be a bit more straightforward?

```suggestion
    # Flag for if the geotransform (of the e.g. HLS tile) crosses the dateline.
    # The dateline crossing by the tile is tested by the condition:
    #     `tile_min_x < 180 and tile_max_x >= 180`
    flag_tile_crosses_dateline = tile_min_x < 180 and tile_max_x >= 180

    # Flag for if the ancillary input file crosses the dateline.
    # The ancillary input only requires reprojection if it is in
    # geographic coordinates and if its longitude domain is represented
    # within the [-180, +180] range, verified by the test `min_x < -170`,
    # rather than the [0, +360] interval.
    flag_file_crosses_dateline = file_srs.IsGeographic() and file_min_x < -170

    # If neither crosses the antimeridian, reproject input file using gdalwarp
    if not (flag_tile_crosses_dateline and flag_file_crosses_dateline):
```

src/proteus/dswx_hls.py
```diff
@@ -3224,11 +3233,17 @@ def _warp(input_file, geotransform, projection,
...
-    # if it doesn't cross the antimeridian, reproject
-    # input file using gdalwarp
-    if not (file_srs.IsGeographic() and tile_min_x < 180 and
-            tile_max_x >= 180):
-
+    # The dateline crossing by the tile is tested by the condition:
+    #     `tile_min_x < 180 and tile_max_x >= 180`
+    # The ancillary input only requires reprojection if it is in
+    # geographic coordinates and if its longitude domain is represented
+    # within the [-180, +180] range, verified by the test `min_x < -170`,
```
nemo794, line 3240: What prompted the decision to use -170 ? Could you please add a comment in the code explaining this? Thanks!

src/proteus/dswx_hls.py
```diff
@@ -4368,7 +4375,15 @@ def _check_ancillary_inputs(check_ancillary_inputs_coverage,
...
-        if file_srs.IsGeographic() and tile_min_x < 180 and tile_max_x >= 180:
+        #
+        # The dateline crossing by the tile is tested by the condition:
+        #     `tile_min_x < 180 and tile_max_x >= 180`
+        # The ancillary input only requires special handling if it is in
+        # geographic coordinates and if its longitude domain is represented
+        # within the [-180, +180] range, verified by the test `min_x < -170`,
+        # rather than the [0, +360] interval.
+        if (file_srs.IsGeographic() and min_x < -170 and
+                tile_min_x < 180 and tile_max_x >= 180):
```
nemo794, line 4386: This looks like code duplication. Please pull this logic out into a standalone function. Please have that function use simplified logic akin to the suggestion in my previous comment, and then call that function in both places.

The dateline crossing issue has been a meaningful concern for SDS et al, and thus might have further tweaks in the future. If so, then it would be extremely easy for a maintainer coming in 5 months from now to update the "check" conditions in one location of the code but not the other, but think they'd completed the bug fix.

Thanks!

gshiroma: Thank you, @nemo794 . Please keep in mind that a new function requires docstrings and retesting. We need a quick fix and make sure everything works.

src/proteus/dswx_hls.py
gshiroma: There's no specific reason to be `-170`. It could be `-160` or even `0`. The important thing is to know that whether the input is using the longitude range [-180, 180] (but obviously we need some margin to test, we cannot use `-180`).

src/proteus/dswx_hls.py
gshiroma: Done

src/proteus/dswx_hls.py
```diff
@@ -4382,21 +4425,21 @@ def _check_ancillary_inputs(check_ancillary_inputs_coverage,
             logger.info(f'    left side (-180 -> +180): {check_1_str}')
 
             # Right side of the antimeridian crossing: +180 -> +360
-            buffer_in_degrees = 0.0002777  # buffer of 1 arcsec: ~ 30m
+            buffer_in_degrees = 33 * 0.0002777  # buffer of 33 arcsec: ~ 1km
```
collinss-jpl, line 4428: I think it'll be handy to assign `33` to a constant near the top of the file so its easy to modify going forward, just in case we find another tile that needs a slightly larger buffer. Same for the `0.0002777`.

collinss-jpl (approved):
Had one suggestion to assign some magic numbers to constants, but otherwise this looks good to me, thanks for the quick work as always @gshiroma!

src/proteus/dswx_hls.py
```diff
@@ -3136,6 +3136,44 @@ def get_projection_proj4(projection):
...
+    ----------
+    file_srs: osr.SpatialReference
+        Ancillary file spatial reference system (SRS)
+    file_min_x: float
+        Ancillary file min longitude value in degrees
+    tile_min_x: float
+        MGRS tile min longitude value in degrees
+    tile_max_x: float
+        MGRS tile max longitude value in degrees
+
```
nemo794, line 3155: Could you please explain what this function returns? Thanks!

src/proteus/dswx_hls.py
```diff
@@ -3155,17 +3155,21 @@ def _antimeridian_crossing_requires_special_handling(
...
-    # within the [-180, +180] range, verified by the test `min_x < -170`,
-    # rather than the [0, +360] interval. There's no specific reason why
-    # -170 is used. It could be -160, or even 0.
+    # Flag to indicate if the if the MGRS tile crosses the antimeridian.
+    flag_tile_crosses_antimeridian = tile_min_x < 180 and tile_max_x >= 180
 
+    # Flag to test if the ancillary input file is in geographic
+    # coordinates and if its longitude domain is represented
+    # within the [-180, +180] range, rather than the [0, +360] interval.
+    # This is verified by the test `min_x < -170`,
```
nemo794, line 3164: Suggestion:
```
    # This is verified by the test `min_x < -170`. There's no specific reason why
    # -170 is used. It could be -160, or even 0.
 ```

src/proteus/dswx_hls.py
gshiroma: Just did it. Thank you, @collinss-jpl

gshiroma:
> Had one suggestion to assign some magic numbers to constants, but otherwise this looks good to me, thanks for the quick work as always @gshiroma!

You provided me with all the info I needed to identify and solve this issue. Thank you so much, @collinss-jpl !

nemo794 (approved):
Thank you for addressing the comments; these changes should help in the long run. I ran a couple of test granules (one that crosses the dateline, the second one not), and both ran to completion. Thanks, @gshiroma !

gshiroma:
> Thank you for addressing the comments; these changes should help in the long run. I ran a couple of test granules (one that crosses the dateline, the second one not), and both ran to completion. Thanks, @gshiroma !

Thank you, @nemo794 , for reviewing the code and testing it over a couple of granules!

Issue #53, opened 2024-01-16
[Bug]: Image Striping - Appears it is related to GDAL Version


cmarshak:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

It appears that when gdal is upgraded to `3.8+` there is issues related to image striping in `PROTEUS`. Firstly, I see this error in the common `dswx_hls.py runconfig.yaml` run e.g. there are several lines in the standard i/o outputting this error:

```
ERROR 1: TIFFSetField:out/OPERA_L3_DSWx-HLS_T53KMA_20210915T011731Z_202401111813Z_1.0/OPERA_L3_DSWx-HLS_T53KMA_20210915T011731Z_202401111813Z_v1.0_B01_WTR.tif: Cannot modify tag "PhotometricInterpretation" while writing
```

This will impact new builds of the software if gdal is not capped or the fix to this error is not found directly in PROTEUS.

I was able to fix this by capping gdal via `mamba install 'gdal<3.8'`.

When I load the data without this cap I see no data "striping" or "banding". Below is a screenshot from a plot without the cap.

![image](https://github.com/nasa/PROTEUS/assets/5975894/d3df9e52-2ace-405b-8d5d-27d67f74bc52)

### What did you expect?

![image](https://github.com/nasa/PROTEUS/assets/5975894/10720088-39a1-49b5-a057-48fcab9646dc)

The above image contains no striping.

### Reproducible steps

```bash
Installing Proteus:

Installing via proteus:

1. `git clone git@github.com:nasa/PROTEUS.git`
2. `conda create --name proteus python=3.9` # haven't tried other python versions yet; this is what is being used by the docker container
3. `conda activate proteus`
4. `conda install --file docker/requirements.txt --yes`
5. `conda install -c conda-forge --file docker/requirements.txt.forge --yes`
7. `python setup.py install`
8. `python setup.py clean`
```

### Environment

```bash
See above.
```

cmarshak:
Use `conda list | grep gdal` to see the version in the relevant environment.

cmarshak:
I created a notebook that tracked this from this tile `HLS.S30.T53KMA.2021258T011731.v2.0`.

Before the `gdal` cap: https://github.com/OPERA-Cal-Val/dswx-hls-pst-workflow/blob/96e6ba869a044aed3d6d533dde1b3cbcd2fa1f70/0-Run-DSWX-HLS.ipynb

[...]

PR #54, opened 2024-01-17
Reorder `SetNoDataValue()` and `WriteArray()` to address issue emerged in GDAL>=3.8

gshiroma:
This PR reorders the setting of the no data value `raster_band.SetNoDataValue()` before the writing of the data array `raster_band.WriteArray()`. This modification aims to address the striping issue (#53) reported by @cmarshak  that emerged when proteus (DSWx-HLS SAS) is executed with GDAL version `3.8` or later.

This PR also updates the proteus version to `v1.0.2` and the release version to `R4.2`


cmarshak:
As an aside, it would be nice to add the problematic HLS tile(s) that we saw this issue into the existing integration test. The issue linked focuses in on`HLS.S30.T53KMA.2021258T011731.v2.0`. However, note when I was trying to figure out this environment issue, this "striping" or "banding" was only present in *certain* tiles. Furthermore, the current test suite passes with `gdal` 3.8+ and *without* this fix. So, this improvement is based on the conversations we are having internally without visibility to the user community. It would be nice to see a test that would have previously failed and now passes. There is likely not time for this, but still putting this out there.

Also, beyond this PR, it would be nice to figure out a way to document the environment that is being used by the production cloud system so when other dependencies invariably get updated, we can figure track down the issue. This reiterates what I was asking @collinss-jpl.

cmarshak:
After thinking about @collinss-jpl description of how SDS stores docker containers (i.e. does an initial build of the container and then recursively uses each subsequent build using the previous container as a base; hopefully getting this roughly correct), it might be helpful longer-term to not necessarily fix the software problem (or other environment issues like it e.g. a newer `gdal` version alters the write API), but rather ensure the software environment is specified in this repository as closely as possible to the production one. Firstly, the changes you made seem correct and look small enough that they would not impact products though honestly, this should be checked on more than just a few tiles. *I think users will be most concerned with generating DSWx-HLS products as closely to those generated by the SDS as possible.*
 
Here are some notes from Scott:

```
I think you’ve hit on our best option for the time being is to capture environment restrictions in the spec/requirements.txt files that are bundled with the repositories. For example, now that we know gdal >= 3.8 is a problem, that could be captured here: https://github.com/nasa/PROTEUS/blob/main/docker/requirements.txt#L1 by changing it to gdal>=3.0,<3.8. We can then reference the spec for tagged releases on the repository to know what went into the container environment. Since the SAS containers all use some variant of a conda environment, part of the dockerfile could also produce a spec file as a build artifact that would capture the exact versions of each package. We could archive those in Jenkins or similar when doing a CI/CD build for a release.
```

RKuttruff (approved):

PR #56, opened 2025-02-20
Update Dockerfile to replace Miniconda with Miniforge

RKuttruff:
Due to anaconda's updated and now strictly enforced licensing requirements, we must replace the miniconda installer with miniforge and ensure all packages are installed from confa-forge or some other non-anaconda channel.

The changes in this PR are documented and evaluated on the OPERA SDS JPL wiki page PGE Subsystem / Switching PGEs from Miniconda to Miniforge


gshiroma (approved):
LGTM. Thank you so much, @RKuttruff !

collinss-jpl:
@gshiroma Thanks for the quick approval, but I'm not sure Riley is able to merge this PR since he does not have write permissions on this repo. Are you able to merge it?

docker/Dockerfile
```diff
@@ -16,15 +16,18 @@ RUN set -ex \
...
+ && conda config --prepend channels conda-forge \
+ && conda config --append channels nodefaults \
+ && conda config --set channel_priority strict \
  && conda config --set show_channel_urls True \
- && conda update --all --yes \
- && conda install --yes --file /tmp/requirements.txt \
- && conda install --yes --channel conda-forge --file /tmp/requirements_forge.txt \
+ && conda update --all --yes --channel conda-forge --override-channels \
+ && conda install --yes --channel conda-forge --override-channels --file /tmp/requirements.txt \
+ && conda install --channel conda-forge --override-channels --file /tmp/requirements_forge.txt \
```
gshiroma, line 30:
```suggestion
```

I don't think we need a different file for conda-forge requirements anymore. Would you mind removing this file (`requirements_forge.txt`) as well? Also, would you please add `seuptools` to `requirements.txt` to see if CI will succeed?

gshiroma (commented):
Thank you again for issuing this PR @RKuttruff . I realize I rushed into approving this PR, but now I think there are some changes we can make. We also need to ensure that the CI is not failing due to this PR.

docker/Dockerfile
gshiroma: There are two other lines in this file, 13 and 31 (before PR updates), that point to `requirements_forge.txt`.If we remove that file, we should remove these lines too.

RKuttruff: I've made these changes, but now I think the issue is with the CI configuration

See #57 - made a PR off of main with no functional code changes and the CI still failed with the same error seen here

gshiroma: Thank you, @RKuttruff ! I see that you added pip install for setuptools. Hopefully, it will work!

RKuttruff: It starts the docker build now
I think I forgot some changes - I pinned some package versions in the requirements.txt

RKuttruff: Ok checks are passing

gshiroma (approved):
LGTM. Thank you so much, @RKuttruff , for issuing this PR replacing miniconda with miniforge and for addressing the CI issue!

RKuttruff:
Happy to help!

